PR #4, opened 2024-10-09
Updated SNR to exclude channels

marcomuzio:
SNR calculation now excludes channels rather than includes them.

### DIFF
--- a/araproc/analysis/snr.py
+++ b/araproc/analysis/snr.py
@@ -97,7 +97,7 @@ def get_snr(waveform):
 
     return snr
 
-def get_avg_snr(wave_bundle, chans=None):
+def get_avg_snr(wave_bundle, excluded_channels=[]):
 
     """
     Calculates channel-wise averaged SNR.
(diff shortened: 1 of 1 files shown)


araproc/analysis/snr.py
```diff
@@ -106,21 +106,22 @@ def get_avg_snr(wave_bundle, chans=None):
...
 
     Returns
     -------
     avg_snr : float
         The average SNR.
     """
 
-    if(chans is None):
-      chans = list(wave_bundle.keys())
+    chans = list(wave_bundle.keys())
```
clark2668, line 118: I don't actually think you have to do this? Doesn't it by default iterate over the keys?

marcomuzio: Ya thats true. Could've cleaned it up a bit more. On the other hand this is probably a bit more explicit what is being done. I'll leave it in for now, can simplify later.

clark2668: Ok! I'll close this for now, it's a really minor nitpick.

PR #6, opened 2024-10-10
Adding first version of RPR Calculation

pgiri4:
Let's make it better !

### DIFF
--- a/araproc/analysis/rpr.py
+++ b/araproc/analysis/rpr.py
@@ -4,6 +4,7 @@ from araproc.framework import waveform_utilities as wfu
 from araproc.analysis import snr
 
 def get_rpr(waveform):
+
     """
     Computes the RPR (Root Power Ratio) value, similar to SNR, for the given waveform.
 
@@ -18,6 +19,7 @@ def get_rpr(waveform):
         The RPR value for the waveform, calculated as the ratio of the maximum voltage 
         (after smoothing) to the standard deviation of the waveform parts with the least RMS.
     """
+
     # Extract time and voltage arrays from the waveform
     channel_time, channel_wf = wfu.tgraph_to_arrays(waveform)
     wf_len = len(channel_wf)
@@ -49,6 +51,7 @@ def get_rpr(waveform):
     return rpr_val
 
 def get_avg_rpr(wave_bundle, chans=None, excluded_channels=[]):
+
     """
     Calculates the average RPR across selected channels from a given set of waveforms.
 
@@ -66,6 +69,7 @@ def get_avg_rpr(wave_bundle, chans=None, excluded_channels=[]):
     avg_rpr: float
         The average RPR value across the selected channels.
     """
+
     chans = list(wave_bundle.keys()) if chans is None else chans
     avg_rpr = []
 
--- a/araproc/analysis/rpr.py
+++ b/araproc/analysis/rpr.py
@@ -1,0 +1,80 @@
+import numpy as np
+from scipy.ndimage import uniform_filter1d
+
+class RPRCalculator:
+    def __init__(self, dt = 1/3.2, sum_win=25,use_debug = False):
+        self.dt = dt
+        self.sum_win = sum_win
+        self.sum_win_idx = int(np.round(self.sum_win / self.dt))
+        self.use_debug = use_debug
+
+    def get_max_info(self):
+        """Find the maximum bin, voltage, and time for the current channel."""
+        self.max_bin = np.nanargmax(self.pad_v)
+        self.max_val = self.pad_v[self.max_bin]
+        self.max_time = self.pad_t[self.max_bin]
+
+        if self.use_debug:
+            print(f"Max value: {self.max_val}, Max time: {self.max_time}")
+
+    def get_sqrt_volt_sum_wf(self):
+        """Apply a rolling mean (smoothing) to the squared waveform."""
+        self.pad_v = np.sqrt(uniform_filter1d(self.pad_v, size=self.sum_win_idx, mode='constant'))
+
+    def get_mean_sigma_in_no_max(self):
+        """Calculate mean and standard deviation excluding the window around the signal peak."""
+        bin_8 = self.pad_num // 8  # Define a window size based on pad_num
+        pad_no_max = self.pad_v.copy()  # Create a copy of the waveform
+        front_idx = max(0,int(self.max_bin - bin_8))
+        
+        if front_idx < 0:
+            front_idx = 0  # Ensure the index doesn't go negative
+        
+        pad_no_max[front_idx:self.max_bin + bin_8 + 1] = np.nan  # Exclude the signal window around the peak
+        
+        self.pad_mean = np.nanmean(pad_no_max)  # Compute mean excluding the peak window
+        self.pad_sigma = np.nanstd(pad_no_max)  # Compute standard deviation excluding the peak window
+
+        if self.use_debug:
+            print(f"Mean: {self.pad_mean}, Sigma: {self.pad_sigma}")
+
+    def get_ch_sliding_v2_snr_uw(self):
+        """Compute the RPR (SNR-like) value."""
+        # Smoothing the waveform (apply rolling mean)
+
+        self.get_sqrt_volt_sum_wf()
+        # Get max bin, value, and corresponding time
+        self.get_max_info()
+
+        # Calculate mean and sigma after excluding the peak window
+        self.get_mean_sigma_in_no_max()
+
+        # Calculate SNR-like (RPR) value
+        self.rpr_arr = (self.max_val - self.pad_mean) / self.pad_sigma
+        
+        # Handle channels where sigma is non-positive
+        nega_sigma_idx = self.pad_sigma <= 0
+        if nega_sigma_idx:
+           print(f'negative std detected, which may be unphysical')
+           self.rpr_arr = -9999 ## Assign some unphysical value
+        
+        if self.use_debug:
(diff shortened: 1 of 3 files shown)


marcomuzio:
Hey @pgiri4, thanks so much for putting this together! This is looking really good, I think it just needs some reorganization. Here are a few changes I would suggest:

- Remove the class structure, I think we let this just be a module of functions which can be directly called
- Try to emulate the structure in snr.py: write a function `get_rpr(waveform)` that calculates the rpr for a single trace then write a function `get_avg_rpr(waveform, excluded_channels)` that averages that result by calling `get_rpr`
- Don't assume things about the trace, like what the timestep is. If you need that information calculate it form the waveform directly. (I think you can safely assume the trace x-axis units are ns. We're not tracking units right @clark2668 ?)
- Add basic documentation for each function following the structure in snr.py

araproc/analysis/rpr.py
```diff
@@ -0,0 +1,206 @@
+import numpy as np
+from scipy.ndimage import uniform_filter1d
+from araproc.framework import waveform_utilities as wfu
+
+
+def get_max_info(channel_wf, channel_time, use_debug=False):
```
marcomuzio, line 6: Do we need the `use_debug` flag? Maybe it's better to simplify and remove the debug statements throughout. If we need to debug we can always add print statements temporarily as needed.

araproc/analysis/rpr.py
```diff
@@ -0,0 +1,206 @@
...
+    max_bin = np.nanargmax(channel_wf)
+    max_val = channel_wf[max_bin]
+    max_time = channel_time[max_bin]
+
+    if use_debug:
+        print(f"Max value: {max_val}, Max time: {max_time}")
+
+    return max_bin, max_val, max_time
+
+def get_sqrt_volt_sum_wf(channel_wf, dt):
```
marcomuzio, line 37: Maybe we can just put this function into `get_ch_sliding_v2_snr_uw`? It's pretty short and only used once.

araproc/analysis/rpr.py
```diff
@@ -0,0 +1,206 @@
...
+    noise_sigma: float
+        Standard deviation excluding the peak window.
+    """
+    bin_8 = wf_len // 8
+    pad_no_max = channel_wf.copy()
+    front_idx = max(0, int(max_bin - bin_8))
+
+    pad_no_max[front_idx:max_bin + bin_8 + 1] = np.nan
+    noise_mean = np.nanmean(pad_no_max)
+    noise_sigma = np.nanstd(pad_no_max)
```
marcomuzio, line 87: Probably we should just replace this function with a call to `snr.get_min_segmented_rms` to get the RMS here? The calculation is similar so we can avoid having a different function. I'm also not sure we need the `noise_mean` anyway.

araproc/analysis/rpr.py
```diff
@@ -0,0 +1,206 @@
...
+    Returns
+    -------
+    rpr_val: float
+        The RPR value (signal-to-noise-like ratio) for the waveform.
+    """
+    dt =  wfu.get_dt_and_sampling_rate(channel_time)[0] 
+    channel_wf = get_sqrt_volt_sum_wf(channel_wf, dt)
+    max_bin, max_val, max_time = get_max_info(channel_wf, channel_time, use_debug)
+    noise_mean, noise_sigma = get_mean_sigma_in_no_max(channel_wf, max_bin, wf_len, use_debug)
+    rpr_val = (max_val - noise_mean) / noise_sigma
```
marcomuzio, line 118: Why is the `noise_mean` subtracted off here? I think the definition in the A23 paper is just `max_val/noise_sigma`

araproc/analysis/rpr.py
```diff
@@ -0,0 +1,206 @@
...
+
+    if noise_sigma <= 0:
```
marcomuzio, line 120: I think this is not possible given the functions you've used. We can keep it if you want but could simplify too.

araproc/analysis/rpr.py
```diff
@@ -0,0 +1,206 @@
...
+
+def get_avg_rpr(wave_bundle, chans=None, individual_antenna=False, use_debug=False):
+    """
+    Calculates channel-wise averaged RPR for a given set of waveforms.
+
+    Parameters
+    ----------
+    wave_bundle: dict of tuples
+        Dictionary containing tuples of (voltage array, time array) for each channel.
+    chans: list, optional
```
marcomuzio, line 164: Can you change this parameter to be `excluded_channels` and make the corresponding changes below.

araproc/analysis/rpr.py
```diff
@@ -0,0 +1,206 @@
...
+        List of channels to average over. If None, averages over all channels in wave_bundle.
+    individual_antenna: bool, optional
```
marcomuzio, line 166: I think we can drop this parameter. If one wants the rpr for a single waveform they should just call run_rpr_calculation for that waveform.

araproc/analysis/rpr.py
```diff
@@ -0,0 +1,206 @@
...
+    """
+    Computes the RPR value for a given waveform by preparing the data and invoking the RPR calculation.
+
+    Parameters
+    ----------
+    Parameters
+    ----------
+    waveform: TGraph
+        A TGraph of the waveform.
+    snr: float, optional
```
marcomuzio, line 139: This parameter doesn't seem to be used, so I think we can drop it.

araproc/analysis/rpr.py
```diff
@@ -0,0 +1,206 @@
...
+    pad_no_max[front_idx:max_bin + bin_8 + 1] = np.nan
+    noise_mean = np.nanmean(pad_no_max)
+    noise_sigma = np.nanstd(pad_no_max)
+
+    if use_debug:
+        print(f"Mean: {noise_mean}, Sigma: {noise_sigma}")
+
+    return noise_mean, noise_sigma
+
+def get_ch_sliding_v2_snr_uw(channel_wf, channel_time, wf_len, use_debug=False):
```
marcomuzio, line 94: Sorry this is a little nitpicky, but can we simplify this function name? Actually given that it returns the rpr itself, maybe this should all be put directly into `run_rpr_calculation`?

araproc/analysis/rpr.py
```diff
@@ -0,0 +1,206 @@
...
+    if noise_sigma <= 0:
+        print(f'Negative std detected, which may be unphysical')
+        rpr_val = -9999
+
+    if use_debug:
+        print(f"RPR: {rpr_val}, Max Val: {max_val}, Mean: {noise_mean}, Sigma: {noise_sigma}")
+
+    return rpr_val
+
+def run_rpr_calculation(waveform, use_debug=False):
```
marcomuzio, line 129: Sorry this is another nitpicky comment, but can we change this function name to `get_rpr`?

pgiri4:
Thanks Marco , thanks Brian. I am happy to have all these feedbacks, I will make changes as suggested.

marcomuzio:
Hey @pgiri4 , thanks for these updates! I think the pull request is much improved. I added some more specific comments to try to simplify things a bit. After those are implemented and reviewed I think it will be good to go!

araproc/analysis/rpr.py
```diff
@@ -0,0 +1,81 @@
...
+    dt =  wfu.get_dt_and_sampling_rate(channel_time)[0]
+    sum_win = 25  # Smoothing window in ns
+    sum_win_idx = int(np.round(sum_win / dt))  # Convert window size to sample points
+    channel_wf = np.sqrt(uniform_filter1d(channel_wf, size=sum_win_idx, mode='constant'))
+
+    # Find the maximum value of the smoothed waveform
+    max_bin = np.nanargmax(channel_wf)
+    max_val = channel_wf[max_bin]
+
+    # Calculate noise sigma: standard deviation of the four parts with the least RMS
```
marcomuzio, line 38: Sorry, I forgot that `snr.get_min_segmented_rms`  didn't accept an array previously. It will now if you do a git pull, so I think we can replace this block by `sigma = snr.get_min_segmented_rms(channel_wf)`.

araproc/analysis/rpr.py
```diff
@@ -0,0 +1,81 @@
...
+    channel_wf = channel_wf ** 2
+
+    # Calculate the smoothing window size based on sampling rate
+    dt =  wfu.get_dt_and_sampling_rate(channel_time)[0]
+    sum_win = 25  # Smoothing window in ns
+    sum_win_idx = int(np.round(sum_win / dt))  # Convert window size to sample points
+    channel_wf = np.sqrt(uniform_filter1d(channel_wf, size=sum_win_idx, mode='constant'))
+
+    # Find the maximum value of the smoothed waveform
+    max_bin = np.nanargmax(channel_wf)
```
marcomuzio, line 35: I think we should remove any `np.nan*` calls throughout. For example, this can just be `np.argmax`. There shouldn't be `nan`'s and if there are we want the code to tell us.

marcomuzio:
@pgiri4 this is FANTASTIC progress!! Really nice job, thank you. I left two quick comments. If we can just address those, it will be done.

araproc/analysis/rpr.py
pgiri4: Hmmm, Okay it was not accepting  the channel_wf before, I can call snr function then.

araproc/analysis/rpr.py
pgiri4: Sure !

marcomuzio:
This looks great, thanks for addressing all my comments!! Merge away @pgiri4 !

PR #7, opened 2024-10-11
Add timing lookup

clark2668:
This adds a small helper function to `araproc.analysis.standard_reco` to lookup the arrival time at an antenna from the ray trace correlator table.

### DIFF
--- a/examples/run_example.py
+++ b/examples/run_example.py
@@ -95,7 +95,7 @@ for e in range(iter_start, iter_stop, 1):
         arrival_time = reco.lookup_arrival_time(channel = 0, 
                                                 theta = reco_results["pulser_v"]["theta"], 
                                                 phi = reco_results["pulser_v"]["phi"],
-                                                which_distance="distant",
+                                                which_distance="nearby",
                                                 solution=0,
                                                 )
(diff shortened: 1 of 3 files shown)


abigailbishop:
I have no comments, this looks great! Thanks Brian!

examples/run_example.py
```diff
@@ -85,11 +85,21 @@
...
+        print(f"The Average SNR is {avg_snr:.1f}")
  
         # run our standard suite of reconstructions
         reco_results = reco.do_standard_reco(wave_bundle)
 
+        # here's how we can lookup arrival times given a reconstructed direction
+        arrival_time = reco.lookup_arrival_time(channel = 0, 
+                                                theta = reco_results["pulser_v"]["theta"], 
+                                                phi = reco_results["pulser_v"]["phi"],
+                                                which_distance="distant",
```
abigailbishop, line 98: I feel like this is a silly question on my side but can you remind me why we're paying pulser reconstructions results with `which_distance="distant"`?

clark2668: Because.... I made a mistake. Fixed now haha! Thank you for catching this!

clark2668:
Thanks @abigailbishop , you caught a bug in my application of the function in the example script! Fixed now, and going to merge.

PR #8, opened 2024-10-14
Add A23 MF_Filters to config file

abigailbishop:


marcomuzio:
Hi @abigailbishop , not sure if this is waiting on anything else on your end. It looks good to me. Maybe my only concern is the filters are pretty narrow in frequency. It wouldn't hurt to widen them slightly I think but maybe they're okay? We will see once we get to processing things!

abigailbishop:
I'm not waiting on anything else but we inherited those filter widths from Myoungchul so I'm happy to start processing data then adjust them later if they seem too strict!

marcomuzio:
Sounds good, I say merge away!

abigailbishop:
Aye captain!

PR #9, opened 2024-10-14
Max correlation and coordinates 

AlanSalcedo:
Added a function to calculate the max correlation and coordinates of a given sky map. I already tested it with:

```
# find the max correlation, and the corresponding theta and phi (also for the vpol map at the distance of the pulser)
max_corr, max_theta, max_phi = correlation.get_max_corr_and_coords(reco_results["pulser_v"]["map"])
print(f"Max correlation: {max_corr}, Theta: {max_theta}, Phi: {max_phi}")
```
Do we want to add the test in run_example.py?

### DIFF
--- a/araproc/analysis/correlation.py
+++ b/araproc/analysis/correlation.py
@@ -1,0 +1,35 @@
+import ctypes
+
+def get_max_corr_and_coords(corr_map):
+    """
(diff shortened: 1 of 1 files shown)


AlanSalcedo:
@marcomuzio I'll close this one and mark the task as done, since we already had it in the reco_results dictionary

PR #10, opened 2024-10-15
Adding A1 CW frequencies

Mohammad-Neutrino:


marcomuzio:
Hey @Mohammad-Neutrino , thanks for adding these! I just have a few small comments. 
1. I think A1 has more than 2 configurations right? You'll need to add a separate section for each configuration. 
2. Are there any excluded channels that shouldn't be used in analysis for any of the configurations? Can you add them to the excluded channel list if there are?
3. For the lower frequency filter, I might raise the threshold to 0.05 to be safe? Since lower frequencies tend to carry more of the power I think, so maybe that's a bit more conservative. That's the level for the A5 filter in that range.

Mohammad-Neutrino:
Hey @marcomuzio, thanks for checking this! 
1.    A1 used to have 4 configs but we decided to drop the outlier configs with trigger window very small (2) and very high (30). Those were mostly either early runs or 2018 not-logged calibration runs. 
2.   A1 does not have any bad channels; only channel 12 and 15 in config 2 were trigger masked which I believe can be useful information for simulation. 
3.   Sure, I can raise the thershold of 225 MHz to be 0.05. 124 MHz CW has low power.

marcomuzio:
Ah right, I forgot there are so few configs now! Okay that all sounds good. Once you've updated the 225 MHz threshold, feel free to merge this. Thanks!

Mohammad-Neutrino:
Added A1 CW frequencies

PR #11, opened 2024-10-15
Surface correlation ratio and coordinates

AlanSalcedo:
Added radius to the reco_results map. Chose to make radii dictionary part of the instantiation in standard_reco.py so I could directly access them in later function.

Wrote a function to calculate surface correlation ratio. The idea is to determine a theta range for the surface region by finding theta for the surface and for surface - z_threshold (need to determine a default for z_threshold). 

Then, find max correlation and coordinates in the surface region bounded by the theta range (all phi accessible). 

Finally divide by the global maximum correlation.


araproc/analysis/correlation.py
```diff
@@ -0,0 +1,114 @@
+import ROOT
+import numpy as np
+import math
+import os
+
+# HELP HERE: constants that we might want to store elsewhere, or can I retrieve them somehow?
```
marcomuzio, line 6: Maybe this is what you are looking for? Not sure if this has a standard reco as part of it already, but if it does you can use that I think. https://github.com/ara-software/AraProc/blob/a758550836aaafa473e05d6d976ee0bf1122dcdf/araproc/analysis/standard_reco.py#L65

clark2668:
Just one comment, but does this need to be it's own thing like `correlation.py`? Seems like this could just be under `standard_reco.py`?

araproc/analysis/correlation.py
```diff
@@ -0,0 +1,114 @@
...
+        The ratio of the maximum surface correlation value within the theta range 
+        to the overall peak correlation.
+    max_surf_theta : float
+        The theta angle corresponding to the maximum surface correlation value.
+    max_surf_phi : float
+        The phi angle corresponding to the maximum surface correlation value.
+    """
+    
+    # Extract the radius from corr_map
+    corr_radius = float(corr_map.get("radius", None))
```
marcomuzio, line 39: Is this the radius of the global maximum? Does this assume a specific map is passed?

araproc/analysis/correlation.py
```diff
@@ -0,0 +1,114 @@
...
+    for ant in range(channel_num):
+        # Retrieve each antenna's location for the given station
+        ant_location = geomTool.getStationInfo(station_id).getAntennaInfo(ant).antLocation
+        z.append(ant_location[2])  # Append only the z-coordinate
+
+    # Calculate the average z-coordinate of the antennas
+    avg_z = sum(z) / len(z)
+
+    # Check if the radius is large enough to visualize the surface
+    if corr_radius < abs(avg_z):
```
marcomuzio, line 57: I think you need `z_thresh` to be part of this

araproc/analysis/correlation.py
```diff
@@ -0,0 +1,114 @@
...
+        raise RuntimeError("Surface not visible: radius is smaller than average z coordinate of the antennas.")
+
+    # Calculate the angles at the surface and z threshold relative to the horizontal
+    theta_at_surface = math.degrees(math.asin(abs(avg_z) / corr_radius))
```
marcomuzio, line 61: I think this should be `acos`?

araproc/analysis/correlation.py
```diff
@@ -0,0 +1,114 @@
+import ROOT
+import numpy as np
+import math
+import os
+
+# HELP HERE: constants that we might want to store elsewhere, or can I retrieve them somehow?
+channel_num = 16
+
+def get_surface_correlation_ratio(corr_map, station_id, z_thresh):
```
marcomuzio, line 9: Can you give `z_thresh` a default value of -10 m? Also we may need to be careful the units of the radius in `corr_map` are the same as `z_thresh`.

araproc/analysis/correlation.py
AlanSalcedo: Yeah, it's the radius of the given map. `corr_map` should be something like `reco_results["distant_v"]` and later we'll catch if the map was created at a large enough radius.

araproc/analysis/correlation.py
AlanSalcedo: `theta` in the map is measured in elevation [90, -90], so I think it's fine?

araproc/analysis/correlation.py
marcomuzio: I wonder if we need pass multiple maps? eg what if the max correlation occurs in a different map than the max surface correlation?

araproc/analysis/correlation.py
marcomuzio: ohhh sorry i thought it was the zenith angle. Okay then I agree you have it right.

araproc/analysis/correlation.py
AlanSalcedo: You'd calculate the surface correlation ratio for that specific map, right? At this point, both the global max and surface corr max are calculated on the same map.

marcomuzio: I'm not sure. For example if the global max across all maps is at the calpulser radius, we still would be interested in knowing the surface correlation ratio. Right now that's not defined. I think we want the surface correlation ratio to be (max surface correlation over all maps)/(max correlation over all maps). Maybe there's a subtlety about the polarizations, but ignoring that for a second I think that's what we want. We can talk about it on the call if you want.

AlanSalcedo: Okay, I can do whatever needed. Just give the order haha. I did it how I understood. 

I will push all the other changes now, including moving this function to standard_reco.py

AlanSalcedo:
Ready for the next check, @marcomuzio :)

araproc/analysis/standard_reco.py
```diff
@@ -330,3 +384,187 @@ def lookup_arrival_time(self,
...
+                corr = hist.GetBinContent(x_bin, y_bin)
+                if theta_min <= theta <= theta_max:
+                    surf_corr_values.append(corr)
+                    corr_bins.append((x_bin, y_bin))
+
+        # Identify maximum surface correlation
+        if not surf_corr_values:
+            raise RuntimeError("No correlation values found in the specified theta range.")
+        
+        max_surf_corr = np.nanmax(surf_corr_values)
```
marcomuzio, line 455: Let's drop the `np.nan*` functions throughout, so this is just `np.max`. If there's a `nan` I think we want there to be an error or something.

araproc/analysis/standard_reco.py
```diff
@@ -330,3 +384,187 @@ def lookup_arrival_time(self,
...
+        return max_surf_corr, max_theta, max_phi
+
+    def get_surface_corr_max_multiple(self, *maps, z_thresh=-10):
+        """
+        Applies get_surface_corr_max to multiple maps and identifies the map with
+        the highest surface correlation value.
+
+        Parameters
+        ----------
+        *maps : dict
```
marcomuzio, line 471: Can you add an example of how to call this? In case the `*maps` thing isn't familiar to folks.

araproc/analysis/standard_reco.py
```diff
@@ -330,3 +384,187 @@ def lookup_arrival_time(self,
...
+        z_coords = [geom_tool.getStationInfo(self.station_id).getAntennaInfo(ant).antLocation[2] 
+                    for ant in range(self.num_channels)]
+        avg_z = sum(z_coords) / len(z_coords)
+
+        # Check if surface is visible at the given radius
+        if radius < (abs(avg_z) + z_thresh):
+            raise RuntimeError("Surface not visible: radius is smaller than average antenna z-coordinate.")
+
+        # Define theta range
+        theta_surface = math.degrees(math.asin(abs(avg_z) / radius))
```
marcomuzio, line 429: Thinking out loud, should we return degrees or radians? I guess the correlation maps are in degrees?

marcomuzio:
It's beautiful @AlanSalcedo !  Just a few small comments. Also looks like there's some conflict with the main branch.

araproc/analysis/standard_reco.py
AlanSalcedo: Degrees. Yes, the corr. maps are in degrees.

araproc/analysis/standard_reco.py
AlanSalcedo: Done!

araproc/analysis/standard_reco.py
AlanSalcedo: Done. Added a ValueError to catch nans.

araproc/analysis/standard_reco.py
marcomuzio: Sounds good!

marcomuzio:
@AlanSalcedo thanks very much for those final changes. Good to merge!

PR #15, opened 2024-10-16
Add CSW and CSW SNR Calculation

abigailbishop:


araproc/analysis/csw.py
```diff
@@ -0,0 +1,479 @@
...
+    final_csw = csw[ start_idx : end_idx+1 ]
+    final_times = big_times[ start_idx : end_idx+1 ]
+
+    # Return requested data
+    if return_rolled_wfs: 
+        return final_times, final_csw, big_times, rolled_wfs
+    else:
+        return final_times, final_csw
+
+def get_csw_snr(csw : np.ndarray):
```
marcomuzio, line 308: I think we can just make this a wrapper around `snr.get_snr(csw)`. That way we are using the same snr definition throughout. :)

abigailbishop: Done!

abigailbishop:
Here's a minimal working example for loading the CSW: 

```
from araproc.framework.dataset import AnalysisDataset
from araproc.analysis import standard_reco as sr
run = "005626"
data = AnalysisDataset(
    station_id = 5,
    path_to_data_file=f"/data/exp/ARA/2019/blinded/L1/ARA05/0701/run{run}/event{run}.root",
    path_to_pedestal_file=f"/data/ana/ARA/ARA05PA/ARA05_pedestals/ped_full_values_A5_run{run}.dat",
    is_simulation = False
)
event = 54
useful_event = data.get_useful_event(event)
wavepacket = data.get_wavepacket(useful_event=useful_event, which_traces='filtered')
reco = sr.StandardReco(data.station_id, excluded_channels=data.excluded_channels)
reco_results = reco.do_standard_reco(wavepacket)['pulser_v']
csw_snr, csw = reco.get_csw_snr(data, useful_event, 0, 0, reco_results, which_distance='nearby', excluded_channels=data.excluded_channels, return_csw=True)
```

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
         avg_hpol_corr_snr = np.mean(avg_hpol_corr_snr)
 
         return avg_vpol_corr_snr, avg_hpol_corr_snr  
+
+    def __get_arrival_delays_reco(
+        self, data, reco_results, channels_to_csw, reference_ch,
```
marcomuzio, line 931: Is there a reason to make this function private? Seems like it could be useful elsewhere.

Also it looks like `data` isn't used. Can we drop that from the argument list?

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+        """
+        Determine the arrival delays from each channel by finding the time of
+        max correlation between each channel and the reference channel in a 
+        `zoom_window` nanosecond window around the expected arrival delay of a 
+        signal given the results from reconstructing the event.
+
+        Parameters
+        ----------
+        wavepacket : dict
+            The wavepacket where keys are not yet channel numbers.
```
marcomuzio, line 993: What does this mean? "where keys are not yet channel numbers"

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+        #   and the reference channel's arrival time
+        reference_arrival_time = arrival_times[reference_ch]
+        arrival_delays = {ch: 0.0 for ch in channels_to_csw }
+        for ch_ID in arrival_delays.keys():
+            arrival_delays[ch_ID] = arrival_times[ch_ID] - reference_arrival_time
+
+        return arrival_delays
+
+    def __get_arrival_delays_AraRoot_xcorr(
+        self, wavepacket, pol, channels_to_csw, reference_ch, reco_delays,
```
marcomuzio, line 981: Looks like `pol` isn't used here. Let's drop it.

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+                    # Software triggers are so short, a channel may not have
+                    #   signal during the time window where signal is expected.
+                    #   As a result, `zoom_indices` will have no entries. 
+                    # Just find the time of peak correlation for the full 
+                    #   waveform in this scenario.
+                    delay = xcorr_times[ np.nanargmax(xcorr_volts) ]
+
+                    # If the event can't find the time window to zoom in on
+                    #   and its not a software trigger, warn user
+                    if trig_type != "soft":
```
marcomuzio, line 1057: This just needs to be switched out.

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+        reference_arrival_time = arrival_times[reference_ch]
+        arrival_delays = {ch: 0.0 for ch in channels_to_csw }
+        for ch_ID in arrival_delays.keys():
+            arrival_delays[ch_ID] = arrival_times[ch_ID] - reference_arrival_time
+
+        return arrival_delays
+
+    def __get_arrival_delays_AraRoot_xcorr(
+        self, wavepacket, pol, channels_to_csw, reference_ch, reco_delays, 
+        trig_type, which_distance, zoom_window=40
```
marcomuzio, line 982: Looks like `which_distance` also isn't used.

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+        # Calculate time difference between each channel's arrival time
+        #   and the reference channel's arrival time
+        reference_arrival_time = arrival_times[reference_ch]
+        arrival_delays = {ch: 0.0 for ch in channels_to_csw }
+        for ch_ID in arrival_delays.keys():
+            arrival_delays[ch_ID] = arrival_times[ch_ID] - reference_arrival_time
+
+        return arrival_delays
+
+    def __get_arrival_delays_AraRoot_xcorr(
```
marcomuzio, line 980: I wonder if we should call this something like `refine_arrival_delays`? My first impression was that this function was just the cross-correlation version of `get_arrival_delays`

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+                    # If the event can't find the time window to zoom in on
+                    #   and its not a software trigger, warn user
+                    if trig_type != "soft": 
+                        print(
+                            f"Touble calculating csw for {trig_type} event with "
+                            f"channels {ch_ID} and {reference_ch}")
+
+                else: 
+                    delay = xcorr_times[ 
+                        np.nanargmax(xcorr_volts[zoomed_indices]) # index of max xcorr in zoomed array
```
marcomuzio, line 1064: Let's just use `np.argmax`?

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+                delay *= -1
+            
+            # Save the calculated arrival delay
+            delays[ch_ID] = delay
+
+        return delays
+
+    def __trim_array(self, times, values, trim):
+        """
+        Trim an array by `trim` values, being careful to avoind trimming out the
```
marcomuzio, line 1081: "Avoind" what a word

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+        times : Iterable
+        values : Iterable
+        """
+
+        # Soft exit if the user doesn't actually request a trim
+        if trim == 0: 
+            return times, values
+        
+        # Identify the index where the peak signal is held
+        value_max_idx = np.nanargmax(values)
```
marcomuzio, line 1101: Just `np.argmax`

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+    def __trim_array(self, times, values, trim):
+        """
+        Trim an array by `trim` values, being careful to avoind trimming out the
+        maximum signal
+
+        Parameters
+        ----------
+        times : Iterable
+        values : Iterable
+        trim : int
```
marcomuzio, line 1088: Is there a reason to do this in terms of index and not trimming to a fixed time window where there will be overlap? Maybe it will be more clear below.

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+        # Determine which channels should be used in the csw
+        if excluded_channels is not None: 
+            if not isinstance(excluded_channels, np.ndarray):
+                excluded_channels = np.array(excluded_channels)
+        else: 
+            excluded_channels = np.array([])
+        channels_to_csw = []
+        for ch_ID in data.rf_channel_indices: 
+            if ((ch_ID not in excluded_channels) and 
+                (data.rf_channel_polarizations[ch_ID] == polarization)):
```
marcomuzio, line 1181: Is this check necessary? I wonder if we should retain the ability to form a csw between polarizations? (If we do drop this we should remove `polarization` as an argument)

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+        reco_results : dict
+            Reco results already with the specific reconstruction reqeusted (so
+            the keys of this object include 'theta' and 'phi')
+        excluded_channels : list
+            Any channels to exclude that aren't already excluded by 
+            livetime configurations or polarization. 
+        which_distance : str
+            The distance of the reconstruction for analysis, following the 
+            convention of other functions in this class. 
+        return_csw : bool
```
marcomuzio, line 1152: We can drop this. It will only return the csw.

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+
+        # Check that reco_results have been passed in properly
+        if 'theta' not in reco_results.keys():
+            raise ValueError(
+                "reco_result object passed into this function does not appear "
+                "to be for a specific reconstruction. Should have a 'theta' key "
+                "but only has the following:", reco_results.keys())
+        
+        # Load wavepacket with waveforms
+        wavepacket= data.get_wavepacket(
```
marcomuzio, line 1169: This is probably slowing it down a little. Maybe `wavepacket` should just be an argument.

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+        """
+        Build the Coherently Summed Waveform (CSW) for the given `useful_event`
+          by shifting each waveform accoriding to the time delay calculated 
+          after finding the max cross correlation between two channels 
+          in a window around the expected arrival delay. 
+
+        Parameters
+        ----------
+        data : AnalysisDataset
+        useful_event : UsefulAtriStationEvent
```
marcomuzio, line 1138: Reminder to remove `useful_event` from arguments and just pass whether its a soft trigger.

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+                "reco_result object passed into this function does not appear "
+                "to be for a specific reconstruction. Should have a 'theta' key "
+                "but only has the following:", reco_results.keys())
+        
+        # Load wavepacket with waveforms
+        wavepacket= data.get_wavepacket(
+            useful_event=useful_event, which_traces='filtered')
+
+        # Determine which channels should be used in the csw
+        if excluded_channels is not None:
```
marcomuzio, line 1173: do we need this? It seems like it just makes sure the list is a np.array? So why not just always do `excluded_channels = np.asarray(excluded_channels)`?

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+            excluded_channels = np.array([])
+        channels_to_csw = []
+        for ch_ID in data.rf_channel_indices: 
+            if ((ch_ID not in excluded_channels) and 
+                (data.rf_channel_polarizations[ch_ID] == polarization)):
+                channels_to_csw.append(ch_ID)
+
+        # In case some channels have different lengths than others, choose the
+        #   smallest waveform size for the length of the csw
+        csw_length = 123456
```
marcomuzio, line 1186: But is it big enough? Maybe `np.inf` is safer?

araproc/analysis/standard_reco.py
clark2668: In which case I would also check the shape of `excluded_channels`.

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+        for ch_ID in channels_to_csw:
+            if wavepacket['waveforms'][ch_ID].GetN() < csw_length: 
+                csw_length = wavepacket['waveforms'][ch_ID].GetN() 
+
+        # Determine the "reference channel" to base the CSW around as the
+        #   channel with the maximum voltage
+        max_voltage = {'voltage': -1, 'channel':-123456}
```
marcomuzio, line 1193: Sorry this is nitpicky, but I think the dictionary overcomplicates it here? If you wanna leave it you can ignore, but we could just track `maxV` and `reference_ch`?

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+        # Get arrival delays relative to the reference channel based on
+        #   expected arrival times from reconstruction results
+        arrival_delays_reco = self.__get_arrival_delays_reco(
+            data, reco_results, channels_to_csw, reference_ch, 
+            which_distance, solution)
+        
+        # Get arrival delays by zooming in on the cross correlation between
+        #   each channel and the reference channel around the expected
+        #   arrival delay from reconstruction results
+        if useful_event.isSoftwareTrigger(): wf_type = 'soft'
```
marcomuzio, line 1210: Sorry can you make a line break so these if/else statements are more readable? (Clearly this is a petpeeve of mine lol)

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+            times = np.asarray(wavepacket['waveforms'][ch_ID].GetX()) - arrival_delays[ch_ID]
+
+            # Determine if the time binning of this channel matches the time binning
+            #   of the CSW and adjust if it doesn't. 
+            # Ex: If the CSW has times of [0.9, 1.4, 1.9] then a channel
+            #   with a time array of [1.6, 2.1, 2.6] needs to be rebinned/shifted 
+            #   backwards by 0.2 but a channel with the time array [3.4, 3.9, 4.4 ] 
+            #   doesn't because each bin starts and ends on the same fraction of a 
+            #   nanosecond that the CSW does. 
+            rebinning_shift = round( (times[0] - csw_times[0]) % csw_dt, 4)
```
marcomuzio, line 1241: Is this going to break if we change our interpolation time step?

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+            # Determine if the time binning of this channel matches the time binning
+            #   of the CSW and adjust if it doesn't. 
+            # Ex: If the CSW has times of [0.9, 1.4, 1.9] then a channel
+            #   with a time array of [1.6, 2.1, 2.6] needs to be rebinned/shifted 
+            #   backwards by 0.2 but a channel with the time array [3.4, 3.9, 4.4 ] 
+            #   doesn't because each bin starts and ends on the same fraction of a 
+            #   nanosecond that the CSW does. 
+            rebinning_shift = round( (times[0] - csw_times[0]) % csw_dt, 4) 
+            if rebinning_shift != 0: 
+                waveform_model = Akima1DInterpolator(times, values)
```
marcomuzio, line 1243: Maybe `waveform_interp` is clearer? But you can ignore.

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+            #   bin isn't about to get rolled 
+            rolled_wf = np.roll( values, int(roll_shift) )
+            rolled_times = np.linspace(
+                times[0] - roll_shift*(times[1] - times[0]),
+                times[-1] - roll_shift*(times[1] - times[0]),
+                len(times)
+            )
+
+            # Add this channel's waveform to the CSW
+            csw_values = np.nansum( np.dstack( (csw_values[0], rolled_wf) ), axis=2)
```
marcomuzio, line 1266: Just `np.sum`?

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,346 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+            rolled_times = np.linspace(
+                times[0] - roll_shift*(times[1] - times[0]),
+                times[-1] - roll_shift*(times[1] - times[0]),
+                len(times)
+            )
+
+            # Add this channel's waveform to the CSW
+            csw_values = np.nansum( np.dstack( (csw_values[0], rolled_wf) ), axis=2) 
+
+        # Un-nest the csw. csw.shape was (1,len(big_times)) but is now len(big_times)
```
marcomuzio, line 1268: I think `big_times` doesn't exist anymore? Also maybe `np.squeeze` is clearer?

araproc/analysis/standard_reco.py
```diff
@@ -3,11 +3,13 @@
 import math
 import os
 import ROOT
+from scipy.interpolate import Akima1DInterpolator
 import itertools
 
 from araproc.analysis import interferometry as interf
 from araproc.framework import constants as const
 from araproc.framework import map_utilities as mu
+from araproc.framework import waveform_utilities as wu
```
marcomuzio, line 12: I think `wfu` is pretty standard in araproc right now, so might be some argument for consistency.

abigailbishop: Done!

araproc/analysis/standard_reco.py
abigailbishop: Done!

araproc/analysis/standard_reco.py
abigailbishop: True, people might call use the arrival delays in other scenarios, I removed the double underscore from this function name and the other arrival delay function

araproc/analysis/standard_reco.py
abigailbishop: I was thinking about this as the cross-correlation version of get_arrival_delays too so I'm leaning towards leaving the name alone but I'm not stubborn about it. If you really want I can change it!

araproc/analysis/standard_reco.py
abigailbishop: Alas it is haha. Fixed

araproc/analysis/standard_reco.py
abigailbishop: Yeah I use this in the context of "waveform 1 is 100 elements long and waveform 2 is 95 elements long, so trim waveform 1 by 5 elements" if that makes sense.

araproc/analysis/standard_reco.py
abigailbishop: The scipy interpolator sometimes uses `nan` values when asked to extrapolate beyond the times it interpolated over. Usually I'd have 1 `nan` value per interpolation at the front or back of a waveform, I can switch to just trimming off the end of the waveforms if you'd prefer that

araproc/analysis/standard_reco.py
abigailbishop: Much better idea than mine. Done!

araproc/analysis/standard_reco.py
abigailbishop: Renamed

araproc/analysis/standard_reco.py
abigailbishop: You're right, I removed the `data` variable. Thanks!

araproc/analysis/standard_reco.py
abigailbishop: Yep thank you!

araproc/analysis/standard_reco.py
abigailbishop: Indeed, thank you!

araproc/analysis/standard_reco.py
abigailbishop: Like the wavepacket has keys of `event` and `waveforms` and the value corresponding to the key of `waveforms` is a dictionary where keys are channel numbers. I wasn't sure how to phrase this comment in a way that would be less likely to need updating if we change the way the wavepackets work in the future. 

Maybe I can say "Wavepacket where keys include waveforms and the event number" if that's better? Or if we're going to be consistent in differentiating 'waveforms' and 'wavepackest' then maybe I'm overthinking this

araproc/analysis/standard_reco.py
abigailbishop: Yep thanks!

araproc/analysis/standard_reco.py
abigailbishop: Agreed, I did that and forgot to push it. Thanks!

araproc/analysis/standard_reco.py
abigailbishop: Done!

araproc/analysis/standard_reco.py
abigailbishop: You're totally right about `big_times`. I have not heard of `np.sqeeze` but I can switch it

araproc/analysis/standard_reco.py
abigailbishop: I changed it to this: "AraProc wavepacket (keys include waveforms and the event number)"

araproc/analysis/standard_reco.py
abigailbishop: Oh trimming to maximize overlap makes way more sense. I'll look into that!

araproc/analysis/standard_reco.py
abigailbishop: Sorry why should we check the shape of `excluded_channels`? What are we checking for?

abigailbishop: Actually I'm removing this whole code block.

araproc/analysis/standard_reco.py
clark2668: So, I think it's a design decision. And so far, we've decided to _exclude_ channels, not tell it what to include. So I'd rather keep that, if that's alright?

abigailbishop: Sure I can revert this commit and switch it to exclusion. The arrival delay functions also work via inclusion, would you like me to switch those to exclusion as well?

araproc/analysis/standard_reco.py
marcomuzio: Ahh right okay. Blerg that used to be configurable but it seems like its the only option now. Maybe the move is right after the interpolation to do a `yint = np.nan_to_num(yint)`. Then we don't have to deal with all these `np.nan*` functions?

araproc/analysis/standard_reco.py
marcomuzio: But is it? I guess in the case that the window in a big number then it's the same. hmm okay up to you!

araproc/analysis/standard_reco.py
abigailbishop: I actually just made a change that has the following logic: If we need to rebin the waveform, we know the first element will be `nan` due to this scipy extrapolation thing so we just trim that single element off of the waveform. I've now switched all the numpy functions with `nan` in the name to not have `nan` in the name. 

If you'd rather, I can tell the scipy interpolator to extrapolate to determine that first datapoint of the waveform.

araproc/analysis/standard_reco.py
abigailbishop: Done!

araproc/analysis/standard_reco.py
abigailbishop: I don't think so? The reconstructed arrival delays are determined by the entire event's estimate of theta and phi whereas this function is determined by finding the time of highest correlation between each channel and a reference. If the time window is too big, the time of highest correlation between two channel could be in a vastly different location of a waveform from the expected arrival window (in the case of near-threshold events) so I'd expect the two arrival delay functions to return different results from each other.

araproc/analysis/standard_reco.py
abigailbishop: The way I've coded this, the first element will always be `nan` if we have to rebin the waveform. I always subtract off the "fraction of a bin shift" so the first element will always be for `original_times[0] - rebinning_shift` which is outside of the interpolation and always `nan`.

Making the waveform 0 I think could bias the RMS of the waveform so I'm favoring just deleting the first bin from the waveform.

araproc/analysis/standard_reco.py
abigailbishop: Done

araproc/analysis/standard_reco.py
abigailbishop: Can you remind me why we would do this? To have a finer CSW? Are we still considering that?

araproc/analysis/standard_reco.py
abigailbishop: I switched all of the functions I've added to work on the basis of exclusion of channels rather and inclusion.

abigailbishop:
Updated version of the minimal working example: 
```
from araproc.framework.dataset import AnalysisDataset
from araproc.analysis import standard_reco as sr
from araproc.analysis import snr
import numpy as np

run = "005626"
data = AnalysisDataset(
    station_id = 5,
    path_to_data_file=f"/data/exp/ARA/2019/blinded/L1/ARA05/0701/run{run}/event{run}.root",
    path_to_pedestal_file=f"/data/ana/ARA/ARA05PA/ARA05_pedestals/ped_full_values_A5_run{run}.dat",
    is_simulation = False
)
event = 54
useful_event = data.get_useful_event(event)
wavepacket = data.get_wavepacket(useful_event=useful_event, which_traces='filtered')
reco = sr.StandardReco(data.station_id, excluded_channels=data.excluded_channels)
reco_results = reco.do_standard_reco(wavepacket)['pulser_v']
csw = reco.get_csw(
    wavepacket, useful_event.isSoftwareTrigger(), 
    0, 0, reco_results, np.concatenate((data.excluded_channels, range(8,17))), 
    which_distance='nearby')
print(snr.get_snr(csw))
```

araproc/analysis/standard_reco.py
abigailbishop: I actually just got rid of the whole coding block that introduces `nan`s and changed all the `np.nan*()` functions to not ignore nan values so this is resolved.

araproc/analysis/standard_reco.py
abigailbishop: Done!

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,364 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+        """
+        Extract the arrival time of the signals from the reconstruction then 
+        calculate the time delays between each channel and the reference channel
+
+        Parameters
+        ----------
+        reco_results : dict
+            Reco results already with the specific reconstruction reqeusted (so
+            the keys of this object include 'theta' and 'phi')
+        rf_channel_IDs : list
```
marcomuzio, line 943: What's the reason for this argument? We don't just wanna pull from `const.rf_channels_ids`?

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,364 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+        # Get arrival delays by zooming in on the cross correlation between
+        #   each channel and the reference channel around the expected
+        #   arrival delay from reconstruction results
+        arrival_delays = self.get_arrival_delays_AraRoot_xcorr(
+            wavepacket, excluded_channels, reference_ch, 
+            arrival_delays_reco, is_software)  
+        
+        # Calculate expected signal time as when the reference channel saw 
+        #   its maximum signal
+        expected_signal_time = wavepacket['waveforms'][reference_ch].GetX()[
```
marcomuzio, line 1169: Can we wrap this `GetX()` in `np.asarray` just in case? If it does somehow create a leak it'll be a pain to find haha. So better to be safe

araproc/analysis/standard_reco.py
```diff
@@ -924,3 +926,364 @@ def get_avg_corr_snr(self, wave_packet, excluded_channels = []):
...
+        for ch_ID in wavepacket['waveforms'].keys(): 
+            if ch_ID not in excluded_channels:
+                channels_to_csw.append(ch_ID)
+
+        # Determine the "reference channel" to base the CSW around as the
+        #   channel with the maximum voltage
+        reference_ch = -123456
+        reference_ch_max_voltage = -1
+        for ch_ID in channels_to_csw:
+            this_max_voltage = np.max(wavepacket['waveforms'][ch_ID].GetY())
```
marcomuzio, line 1149: also this one, wrap it in `np.asarray`. I'm too paranoid lol

marcomuzio:
@abigailbishop okay I'm set. I left a few small comments, but I think this is good to go really!

araproc/analysis/standard_reco.py
abigailbishop: I did not know that that existed haha. Updated!

araproc/analysis/standard_reco.py
abigailbishop: Sure! Anything with .GetN should be fine right?

araproc/analysis/standard_reco.py
abigailbishop: Yep, done!

araproc/analysis/standard_reco.py
abigailbishop: Roger that thanks. It's been fixed!

abigailbishop:
Thanks so much Marco!

PR #16, opened 2024-10-16
Refracted/Reflected reconstruction maps

AlanSalcedo:
Added interferometric maps for refracted/reflected solutions for distant solutions.


araproc/analysis/standard_reco.py
```diff
@@ -236,34 +243,81 @@ def do_standard_reco(self, waveform_bundle):
...
+            "radius": self.distant_events_r_library[1]
+        }
+
+        # Store the refracted/reflected rays results
+        reco_results["distant_v_ref"] = {
+            "corr": corr_distant_v_ref, 
+            "theta": theta_distant_v_ref,
+            "phi": phi_distant_v_ref,
+            "map": distant_map_v_ref,
+            "radius": self.distant_events_r_library[1]
```
clark2668, line 283: I would prefer if you did `self.rtc_wrapper.correlators["distant"].GetRadius()`

clark2668: That way the only thing that's important is that you have your handle on the right correlator. It's more modular (I think...)

clark2668 (changes requested):

araproc/analysis/standard_reco.py
AlanSalcedo: Ah, I didn't know I could get the radius from there. Thanks!

clark2668: Yup! :)

AlanSalcedo: Done. Should I merge or wait for @marcomuzio?

marcomuzio: Merge away!

PR #17, opened 2024-10-16
Add support for np.ndarrays for snr functions.

marcomuzio:
Added support for np.ndarrays to be passed to snr functions rather than just TGraphs.

### DIFF
--- a/araproc/analysis/snr.py
+++ b/araproc/analysis/snr.py
@@ -11,7 +11,7 @@ def get_vpp(waveform):
     Parameters
     ----------
     waveform: TGraph or np.ndarray
-        A TGraphi or np.ndarray of the waveform voltage.
+        A TGraph or np.ndarray of the waveform voltage.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


araproc/analysis/snr.py
```diff
@@ -9,16 +10,21 @@ def get_vpp(waveform):
...
     Returns
     -------
     vpp : float
         Peak-to-peak voltage in same units as trace.
     """ 
 
-    _, trace = wfu.tgraph_to_arrays(waveform)
+    if(isinstance(waveform, ROOT.TGraph)):
+      _, trace = wfu.tgraph_to_arrays(waveform)
+    elif(isinstance(waveform, np.ndarray)):
```
clark2668, line 24: I actually think you want to be more aggressive than this. You don't just care that it's an ndarray. For this to make sense, it needs to be an ndarray of a specific shape, in particular it needs to be a 1D array. So I'd add enforcement for that.

clark2668 (changes requested):

araproc/analysis/snr.py
marcomuzio: Okay how's that?

PR #18, opened 2024-10-17
fixing a small A1 stationID bug

Mohammad-Neutrino:
There was a minor A1 stationID issue in _standard_reco.py_ while loading timing tables using _calpulser_r_library_


clark2668:
Yup, looks good. Merge away.

PR #19, opened 2024-10-17
 Here is a draft of daq error and read out window error cut

pgiri4:


marcomuzio:
Hi @pgiri4, thanks for getting this together. I think it just needs some restructuring and it will be there. Here are a few broad comments that should help get it most of the way:

1. Assume the input for functions is `useful_event` from `useful_event = data.get_useful_event()`. These functions only need to do these quality checks for a single event at a time (the event loaded into `useful_event`). 
2. Since we're only checking things for a single event, you can use the pyroot way of accessing variables. For example, to get the event `timeStamp` you would do `useful_event.timeStamp`. 
3. Use AraRoot functions where they exist. For example to check if the event is a calpulser you can use `useful_event.isCalpulserEvent()`. Corresponding functions can be found [here](https://github.com/ara-software/AraRoot/blob/a86445f219bbcb57dd8b4d5f49332a7c5b5d02b3/AraEvent/RawAtriStationEvent.h#L69).
4. For information that can't be gotten from `useful_event` and needs to be supplied in some way by the user, we can add to `AraProc/araproc/framework/config_files/analysis_configs.yaml` and ask everyone to add the necessary info. Then you can code it assuming you pull the relevant info from that file. An example of how to do than can be found [here](https://github.com/ara-software/AraProc/blob/85281a48dbda6bf366434c2d00e98fa2b3713441/araproc/framework/dataset.py#L53).
5. The function that gets called by the user should be something like `check_daq_quality(useful_event)` which is a wrapper for all the other functions you've written and returns True or False to indicate if all the quality checks have passed.

Let us know if you have questions or run into errors implementing any of these, so we can get this turned around quickly.

araproc/analysis/daq_quality_cut.py
```diff
@@ -0,0 +1,265 @@
...
+    block_nums = np.empty(num_blocks)  # Create an empty array for block numbers
+    channel_mask = np.empty(num_blocks, dtype=int)  # Create an empty array for channel masks
+    for n in range(num_blocks):
+        this_block_id = useful_event.blockVec[n].getBlock()
+        block_nums[n] = int(this_block_id)
+        channel_mask[n] = useful_event.blockVec[n].channelMask
+    # Extract event-level information
+    read_win = useful_event.numReadoutBlocks  # Number of readout blocks
+
+    trig_type = 0 if useful_event.isRFTrigger() else (1 if useful_event.isCalpulserEvent() else (2 if useful_event.isSoftwareTrigger() else -1))
```
marcomuzio, line 43: Can you break this up into a few lines so it's more readable?

araproc/analysis/daq_quality_cut.py
```diff
@@ -0,0 +1,265 @@
...
+
+    num_ddas = 4
```
marcomuzio, line 45: I think some numbers like this are stored in araproc/framework/constants.py

araproc/analysis/daq_quality_cut.py
```diff
@@ -0,0 +1,265 @@
...
+    channel_mask : array-like
+        The channel mask for the event.
+
+    Returns
+    -------
+    bool
+        True if there are DAQ structure errors, False otherwise.
+    """
+
+    num_ddas = 4  # Number of DDAs (Data Digitization Assemblies)
```
marcomuzio, line 73: These constants too.

araproc/analysis/daq_quality_cut.py
```diff
@@ -0,0 +1,265 @@
...
+    return np.sum(daq_st_err) > 0
+
+
+def get_read_win_limit(st, run):
+
+    """
+    Determine the readout window limits based on station and run number.
+    Returns the RF and soft readout limits.
+    """
+    if st == 1:
```
marcomuzio, line 132: Can we add these to the yaml configuration files? And read from there?

araproc/analysis/daq_quality_cut.py
```diff
@@ -0,0 +1,265 @@
...
+    useful_event : object
+        The useful event object containing event information.
+    station_id : int
+        The station ID for the event.
+    run : int
+        The run number for the event.
+
+    Returns
+    -------
+    daq_errors : bool
```
marcomuzio, line 249: I really like this function! I think we can form an `or` between the two bools listed here and just return a single bool to make clear if there are any errors.

marcomuzio:
@pgiri4 this is a huge improvement! Thanks for addressing those big restructuring comments. I left a few small comments, but I think this is just about done.

araproc/analysis/daq_quality_cut.py
```diff
@@ -0,0 +1,253 @@
...
+    block_nums = np.empty(num_blocks)  # Create an empty array for block numbers
+    channel_mask = np.empty(num_blocks, dtype=int)  # Create an empty array for channel masks
+    for n in range(num_blocks):
+        this_block_id = useful_event.blockVec[n].getBlock()
+        block_nums[n] = int(this_block_id)
+        channel_mask[n] = useful_event.blockVec[n].channelMask
+    # Extract event-level information
+    read_win = useful_event.numReadoutBlocks  # Number of readout blocks
+
+    if useful_event.isRFTrigger():
```
marcomuzio, line 43: I can't remember if calpulsers are also rf triggers on ARA. They are on PA in which case the order here matters. @clark2668 do you know off the top of your head?

araproc/analysis/daq_quality_cut.py
```diff
@@ -0,0 +1,253 @@
...
+       trig_type = 0
+    elif useful_event.isCalpulserEvent():
+       trig_type = 1
+    elif useful_event.isSoftwareTrigger():
```
marcomuzio, line 47: Probably should add an `else` and throw an error? so you don't get an undefined behavior

araproc/analysis/daq_quality_cut.py
```diff
@@ -0,0 +1,253 @@
...
+        Run Number
+    Returns
+    -------
+    rf_readout_limit: int
+        RF trig readout window limit.
+    soft_readout_limit: int
+        Soft trig readout window limit.
+    """
+
+    file = pkg_resources.open_text(config_files,"analysis_configs.yaml")
```
marcomuzio, line 152: This is awesome! Can you update the yaml files with some placeholder values for folks to update?

araproc/analysis/daq_quality_cut.py
clark2668: Yup! That is correct. Cal pulsers _are_ RF triggers. They can be "tagged" based on their timing under normal operating conditions. But they fundamentally are RF triggers.

pgiri4: Hmm. So, the option would be something like this, right or I should consider sth else as well? 
image

araproc/analysis/daq_quality_cut.py
pgiri4: Haha, the irony is I thought I pushed the yaml file as well.

araproc/analysis/daq_quality_cut.py
marcomuzio: Ya you can do it like that or like:
```
if useful_event.isCalpulserEvent():
  trig_type = 1
elif useful_event.isRFTrigger():
  trig_type = 0
```

araproc/analysis/daq_quality_cut.py
marcomuzio: haha no worries! let's just make sure to add it before merging :)

araproc/framework/config_files/analysis_configs.yaml
```diff
@@ -14,6 +14,17 @@ station100:
       filt3 : { min_freq : 0.396, max_freq : 0.410, min_power_ratio : 0.02 }
       filt4 : { min_freq : 0.570, max_freq : 0.580, min_power_ratio : 0.02 }
 
+  readout_limits:
```
marcomuzio, line 17: I think these are config dependent? So they probably need to be indented and put into each configuration?

marcomuzio:
@pgiri4 thank you so much for your work here, really nice job! Merge away! :)

PR #20, opened 2024-10-17
Minimum correlation fraction depth

AlanSalcedo:
This function calculates the minimum correlation fraction depth:

You can specify (1) a fraction of the peak correlation -default 60%- and (2) a threshold depth where the surface "starts" or "it's reached" - neg 10 m- by default. 

This function starts a 'min_depth' from a neg. infinity depth and replaces 'min_depth' when it finds a new depth (measured from the surface and not above the surface) with a correlation => max_corr * fraction.

### DIFF
--- a/araproc/analysis/standard_reco.py
+++ b/araproc/analysis/standard_reco.py
@@ -625,8 +625,6 @@ class StandardReco:
         # Get maximum correlation and threshold
         max_corr = float(corr_map.get("corr"))
         threshold_corr = fraction * max_corr
-        print("Max corr in map: ", max_corr)
-        print("Threshold corr: ", threshold_corr)
 
         # Access correlation map and radius
         hist = corr_map.get("map")
@@ -655,7 +653,6 @@ class StandardReco:
                     # Ensure min_depth does not exceed z_thresh
                     if depth > min_depth and depth <= z_thresh:
                         min_depth = depth
-                        print(f"Updated min_depth to {min_depth} Theta: {theta}, Corr: {corr_value}, Depth: {depth}")
 
         if min_depth == -float('inf'):
             raise RuntimeError("No depth meets the fractional correlation threshold within the z_thresh.")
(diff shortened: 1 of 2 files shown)


araproc/analysis/standard_reco.py
```diff
@@ -599,3 +599,61 @@ def calculate_surface_corr_ratio(self, *maps, z_thresh=-10):
         
         return surf_corr_ratio, max_surf_corr_result, max_corr_result
 
+    def min_frac_corr_depth(self, corr_map, fraction=0.6, z_thresh=-10):
```
marcomuzio, line 602: I think this is another one we want to aggregate over maps, same as with the surface correlation ratio. Also if we have `z_thresh` let's make the default value `0`.

araproc/analysis/standard_reco.py
```diff
@@ -599,3 +599,61 @@ def calculate_surface_corr_ratio(self, *maps, z_thresh=-10):
...
+        threshold_corr = fraction * max_corr
+
+        # Access correlation map and radius
+        hist = corr_map.get("map")
+        radius = corr_map.get("radius")
+        if hist is None or radius is None:
+            raise ValueError("Map or radius not found in corr_map.")
+        radius = float(radius)
+
+        # Calculate average antenna z-coordinate
```
marcomuzio, line 636: Now that I've seen this a few times, I wonder if we should just make this its own function.

araproc/analysis/standard_reco.py
```diff
@@ -599,3 +599,61 @@ def calculate_surface_corr_ratio(self, *maps, z_thresh=-10):
...
+        fraction : float
+            Fraction of the max correlation to set the threshold.
+            Default is 0.6 (or 60%).
+        z_thresh : float
+            Maximum allowable depth (relative to the surface). Default is -10 meters.
+
+        Returns
+        -------
+        min_depth : float
+            Shallowest depth where correlation meets the specified fraction of max, adjusted by avg z, and below z_thresh.
```
marcomuzio, line 623: Maybe this can be more clear. Just "Shallowest depth (measured from ice surface) where correlation meets the specified fraction of max and below z_thresh."

araproc/analysis/standard_reco.py
marcomuzio: Seems like it could be at home in araproc/framework/map_utilities.py?

AlanSalcedo:
@marcomuzio ready for another check :)

araproc/analysis/standard_reco.py
marcomuzio: Sorry I think we want this one to have a `*maps` arguments right?

AlanSalcedo: We do that in the "multiple" version: 'min_frac_corr_depth_multiple(self, *maps, fraction=0.6, z_thresh=0):' where each element of *maps is passed to 'min_frac_corr_depth(self, corr_map, fraction=0.6, z_thresh=-10)' and then identifies which map has minimum depth

marcomuzio: Whoops! Sorry I missed that. Awesome, then it's all set!

marcomuzio:
@AlanSalcedo merge that thang

PR #21, opened 2024-10-17
correlation_snr, hilbert_envelope_snr, and spark_power_ratio calculations

Mohammad-Neutrino:


araproc/analysis/correlation_snr.py
```diff
@@ -0,0 +1,254 @@
...
+import scipy
+import itertools
+import numpy as np
+from scipy.signal import correlate as co
+
+from araproc.analysis import snr
+from araproc.framework import waveform_utilities as wfu
+
+
+def channel_pairs():
```
marcomuzio, line 13: I think to keep things simple, we'll want to put all of your correlation_snr.py into standard_reco.py. A lot of these functions already exist there and are precalculated.

araproc/analysis/correlation_snr.py
```diff
@@ -0,0 +1,254 @@
...
+
+    pairs = list(itertools.combinations(rf_chs, 2))
+    pairs_V = list(itertools.combinations(rf_chs[:8], 2))
+    pairs_H = list(itertools.combinations(rf_chs[8:], 2))
+
+    return pairs, pairs_V, pairs_H
+
+
+
+def pairing(channels):
```
marcomuzio, line 42: I think we'll use the pairing scheme in standard_reco.py

araproc/analysis/correlation_snr.py
```diff
@@ -0,0 +1,254 @@
...
+    -------
+    corr: array
+       Array of values of cross correlation function.  
+    """
+
+    _, y1 = wfu.tgraph_to_arrays(y1)
+    _, y2 = wfu.tgraph_to_arrays(y2)
+    
+    n = len(y1)
+    corr = co(y2, y1, mode = 'same')/np.sqrt(co(y1, y1, mode = 'same')[int(n/2)]*co(y2, y2, mode = 'same')[int(n/2)])
```
marcomuzio, line 86: You'll pull this directly from what's been stored in standard_reco.py

araproc/analysis/correlation_snr.py
```diff
@@ -0,0 +1,254 @@
...
+    corr_snr, VPol_corr_snr, HPol_corr_snr = {}, {}, {}
+
+    for chan in chans:
+
+      if(chan in excluded_channels):
+        continue
+
+      AllPol_dict.update({chan: wave_bundle[chan]})
+
+      if chan in chans[:8]:
```
marcomuzio, line 151: we should check the channel numbers, what if chans doesn't include the channels you expect? or in the right order?

araproc/analysis/correlation_snr.py
```diff
@@ -0,0 +1,254 @@
...
+      if chan in chans[8:]:
+        HPol_dict.update({chan: wave_bundle[chan]})
+    
+    avg_corr_snr, avg_vpol_corr_snr, avg_hpol_corr_snr = [], [], []
+
+    pairs = pairing(list(AllPol_dict.keys()))
+    pairs_V = pairing(list(VPol_dict.keys()))
+    pairs_H = pairing(list(HPol_dict.keys()))
+
+    for pair in pairs:
```
marcomuzio, line 162: May not matter once we switch to the pairs in standard_reco.py, but we should be careful pairs aren't double counted.

araproc/analysis/hilbert_envelope_snr.py
```diff
@@ -0,0 +1,110 @@
...
+
+import numpy as np
+from scipy.signal import hilbert
+
+from araproc.analysis import snr
+from araproc.framework import waveform_utilities as wfu
+
+
+
+def get_envelope(waveform, nsegs = 8):
```
marcomuzio, line 11: I think we can drop `nsegs` from the arguments.

araproc/analysis/hilbert_envelope_snr.py
```diff
@@ -0,0 +1,110 @@
...
+    waveform: TGraph
+        A TGraph of the waveform.
+
+    Returns
+    -------
+    hill_snr : float
+        The Hilbert envelope SNR of the waveform.
+    """
+
+    hill_max, hill_rms = get_envelope(waveform)
```
marcomuzio, line 67: do we need a separate `get_envelope` function? Seems like this is simple enough to put into one function.

araproc/analysis/spark_power_ratio.py
```diff
@@ -0,0 +1,102 @@
...
+
+import numpy as np
+
+from araproc.framework import waveform_utilities as wfu
+
+
+
+
+
+def get_rms_power(waveform):
```
marcomuzio, line 12: Is this the rms power? Maybe its more like the mean power?

araproc/analysis/spark_power_ratio.py
```diff
@@ -0,0 +1,102 @@
...
+        A TGraph of the waveform.
+
+    Returns
+    -------
+    power: float
+        RMS power of the waveform.
+    """
+
+    _, trace = wfu.tgraph_to_arrays(waveform)
+    power = np.nanmean(trace**2)
```
marcomuzio, line 29: we should drop the `np.nan*` functions. We want it to complain if a `nan` appears. So this can just be `np.mean`

araproc/analysis/spark_power_ratio.py
```diff
@@ -0,0 +1,102 @@
...
+
+    return power
+
+
+def threshold(threshold_ratio = 5.0):
```
marcomuzio, line 34: I think we can drop this function altogether. Thresholds will be set and assessed in post-processing.

araproc/analysis/spark_power_ratio.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    Calculates power ratio between largest and second largest rms power strings.
+
+    Parameters
+    ----------
+    wave_bundle: dict of TGraphs
+        Dictionary of waveform TGraphs to be averaged.
+    excluded_channels: list
+        List of dictionary keys to exclude from average.
+    cut_ratio: float
```
marcomuzio, line 66: Can drop this argument

araproc/analysis/spark_power_ratio.py
```diff
@@ -0,0 +1,102 @@
...
+        Power ratio threshold.
+
+    Returns
+    -------
+    power_ratio: float
+        Power ratio of largest to second largest rms power strings.
+    is_spark_event: bool
```
marcomuzio, line 73: Let's just return the `power_ratio`

araproc/analysis/spark_power_ratio.py
```diff
@@ -0,0 +1,102 @@
...
+        0: is not a spark event, 1: is a spark event.
+    """
+
+    if cut_ratio is not None:
+       cut_ratio = float(cut_ratio)
+    else:
+       cut_ratio = threshold()
+
+
+    num_strings = 4
```
marcomuzio, line 83: I think this info is stored somewhere in AraProc. We should make this work if there are fewer strings than expected

araproc/analysis/spark_power_ratio.py
```diff
@@ -0,0 +1,102 @@
...
+
+    chans = list(wave_bundle.keys())
+    string_chans = [chans[i::4] for i in range(num_strings)]
```
marcomuzio, line 86: Careful here about whether the channels are ordered properly. Probably better to check the string number with something like `strNo = chans[i]%4`

araproc/analysis/spark_power_ratio.py
```diff
@@ -0,0 +1,102 @@
...
+
+    power = []
+    for ch in np.array(string_chans).flatten():
+       if ch in excluded_channels:
+          power.append(np.nan)
```
marcomuzio, line 91: I think better to just `continue`?

marcomuzio:
@Mohammad-Neutrino thanks a lot and nice work! Overall I think this is in really good shape. You've got the right philosophy and structure for things. Just left some comments to simplify and improve in a few places.

araproc/analysis/correlation_snr.py
Mohammad-Neutrino: Does that mean, we are not keeping a dedicated correlation_snr.py script?

araproc/analysis/correlation_snr.py
Mohammad-Neutrino: Right! So removing 

`if(chan in excluded_channels):
        continue`
        
would solve the issue. Later excluded channels can be discarded. As far channel mapping, the RF channels are same for all stations, no? I thought, when you call getGraphFromRFChan, it loads the correctly mapped Electric Channels.

araproc/analysis/hilbert_envelope_snr.py
Mohammad-Neutrino: I kept 'nsegs' in case we want to use the commented portion script. But yeah, I can move that inside the commented part.

araproc/analysis/spark_power_ratio.py
Mohammad-Neutrino: I call it RMS power because you get this technically by squaring the RMS voltage. But yes, this is basically mean power. I can change the name if required!

araproc/analysis/spark_power_ratio.py
Mohammad-Neutrino: sure thing, I can fix that.

araproc/analysis/spark_power_ratio.py
Mohammad-Neutrino: Sure thing, just kept it there for future use in case we wanted to set different threshold for separate stations and per configurations. But I can remove it or keep it and terminate its dependencies on other functions.

araproc/analysis/spark_power_ratio.py
Mohammad-Neutrino: That's fair. let me look for that

araproc/analysis/spark_power_ratio.py
Mohammad-Neutrino: continue will complain when you average an inhomogeneous array of power. so appending `nan` and later `nanmean ` makes it work.

Mohammad-Neutrino:
Thanks a lot, @marcomuzio , I have left some responses. Let me know what you think or we can talk about this in coding session today. And then decide what to do. Thanks again!

araproc/analysis/correlation_snr.py
marcomuzio: Right, I think in this case we don't need it.

araproc/analysis/correlation_snr.py
marcomuzio: Sorry i mean more what is the first 8 entries in `chans` aren't vpols? like what if `chans=[1, 14, 3, 13, ...]` for example

araproc/analysis/hilbert_envelope_snr.py
marcomuzio: I think a `get_envelope` function would make sense if it returns the Hilbert envelope itself. With that we could easily do what you are saying and also have the benefit of being able to, eg, plot it. Otherwise, I think it can be merged.

araproc/analysis/spark_power_ratio.py
marcomuzio: Hm, but then its actually squared rms voltage? Which is the mean power not the rms power? I think we should just make sure the name reflects what's actually calculated so people don't grab the function thinking it calculates one thing when it's calculating another.

araproc/analysis/spark_power_ratio.py
marcomuzio: Ya I think that will be done elsewhere, probably better to remove it for now so we don't accidentally implement the threshold at this point.

araproc/analysis/spark_power_ratio.py
marcomuzio: I'm not sure the keys in a python dictionary are guaranteed to be ordered.

marcomuzio: Looks like you need an `OrderedDict` to be sure. Otherwise I think it's supposed to be ordered based on the order you added to the dictionary? So no guarantees.

https://stackoverflow.com/questions/5629023/order-of-keys-in-dictionaries-in-old-versions-of-python

araproc/analysis/spark_power_ratio.py
marcomuzio: Hmm I'm not sure I understand that. Is it becuz you are forming a 2d array that would then be inhomogeneous. Probably then you should do something like:
```
meanPower = []
power = defaultdict(list)
for strNo in range(noStrs):
  for ch in strChans:
    thisPower = [calculate power of chan ch]
    power[strNo].append(thisPower)
  meanPower.append(np.mean(power))
```

araproc/analysis/hilbert_envelope_snr.py
marcomuzio: Ah I see now. I think the commented part isn't going to be used at this point, so probably better to remove the comment block too to make it less cluttered.

araproc/analysis/spark_power_ratio.py
Mohammad-Neutrino: And that should be done while generating those wave_bundle, right?

araproc/analysis/spark_power_ratio.py
Mohammad-Neutrino: sure, this works as well but where is it taking care of excluded channels? Also, I always try to avoid multiple loops whenever I can but nan can be dangerous too.

araproc/analysis/spark_power_ratio.py
marcomuzio: i mean, i dont think we can say how they are generated. we should just not assume anything about the ordering.

araproc/analysis/spark_power_ratio.py
marcomuzio: you can just add a `continue` to the inner loop
```
if(ch in excluded_chans):
  continue
```

araproc/analysis/hilbert_envelope_snr.py
```diff
@@ -0,0 +1,110 @@
+# this calculates Hilbert envelope SNR for each channel and their average
```
marcomuzio, line 1: sorry this is so nitpicky. can we change the file name to just hilbert.py. i am liking the idea of it providing the utility to calculate the hilbert envelope so others can call it.

araproc/analysis/hilbert.py
```diff
@@ -0,0 +1,97 @@
...
+    excluded_channels: list
+        List of dictionary keys to exclude from average.
+
+    Returns
+    -------
+    avg_hill_snr : float
+        The average Hilbert envelope SNR.
+    """
+
+    chans = sorted(list(wave_bundle.keys()))
```
marcomuzio, line 84: I'm not sure the sort is necessary here, so could drop it.

araproc/analysis/standard_reco.py
```diff
@@ -701,3 +704,135 @@ def min_frac_corr_depth_multiple(self, *maps, fraction=0.6, z_thresh=0):
...
+        -------
+        pairs: list
+           List of all possible pair combinations of given channels.
+        """
+
+        pairs = list(itertools.combinations(channels, 2))
+
+        return pairs
+
+    def get_corr(y1, y2):
```
marcomuzio, line 728: We can drop this function. You'll use `__get_correlation_function(ch1, ch2, event_num)` to get the correlation TGraph.

araproc/analysis/standard_reco.py
```diff
@@ -701,3 +704,135 @@ def min_frac_corr_depth_multiple(self, *maps, fraction=0.6, z_thresh=0):
...
+
+        _, y1 = wfu.tgraph_to_arrays(y1)
+        _, y2 = wfu.tgraph_to_arrays(y2)
+
+        n = len(y1)
+        corr = co(y2, y1, mode = 'same')/np.sqrt(co(y1, y1, mode = 'same')[int(n/2)]*co(y2, y2, mode = 'same')[int(n/2)])
+
+        return corr
+
+    def get_corr_snr(first_ch, second_ch):
```
marcomuzio, line 754: These functions need to be member functions so should have a first argument of `self` that doesn't get used. So this would be `def get_corr_snr(self, first_ch, second_ch)`.

araproc/analysis/standard_reco.py
```diff
@@ -701,3 +704,135 @@ def min_frac_corr_depth_multiple(self, *maps, fraction=0.6, z_thresh=0):
...
+        corr_snr: float
+           Channel pair correlation SNR.
+        """
+
+        corr = get_corr(first_ch, second_ch)
+        corr_snr = snr.get_snr(corr)
+
+        return corr_snr
+
+    def get_avg_corr_snr(wave_bundle, excluded_channels = []):
```
marcomuzio, line 777: Same here, will need a `self` argument. Also will need to take an `event_number` argument for technical reasons.

araproc/analysis/standard_reco.py
```diff
@@ -701,3 +704,135 @@ def min_frac_corr_depth_multiple(self, *maps, fraction=0.6, z_thresh=0):
             raise RuntimeError("No maps meet the fractional correlation threshold within the z_thresh.")
 
         return min_depth, min_depth_index
+
+    def pairing(channels):
```
marcomuzio, line 708: I think maybe this isn't needed anymore.

araproc/analysis/standard_reco.py
```diff
@@ -701,3 +704,135 @@ def min_frac_corr_depth_multiple(self, *maps, fraction=0.6, z_thresh=0):
...
+            if chan in const.hpol_channel_ids:
+                HPol_dict.update({chan: wave_bundle[chan]})
+
+        avg_corr_snr, avg_vpol_corr_snr, avg_hpol_corr_snr = [], [], []
+
+        pairs = pairing(list(AllPol_dict.keys()))
+        pairs_V = pairing(list(VPol_dict.keys()))
+        pairs_H = pairing(list(HPol_dict.keys()))
+
+        for pair in pairs:
```
marcomuzio, line 822: I think you can just iterate over the pre-made vpol and hpol pairs. You can access them as `self.pairs_v` and `self.pairs_h`. Take a look at the `get_pair_index` function (part of the open PR) for an example of how this works.

araproc/analysis/hilbert.py
```diff
@@ -0,0 +1,97 @@
...
+
+import numpy as np
+from scipy.signal import hilbert
+
+from araproc.analysis import snr
+from araproc.framework import waveform_utilities as wfu
+
+
+
+def get_envelope(waveform):
```
marcomuzio, line 11: Sorry can this function be moved to `AraProc/araproc/framework/waveform_utilities.py`?

araproc/analysis/spark_power_ratio.py
```diff
@@ -0,0 +1,71 @@
...
+    
+    Returns
+    -------
+    power_ratio: float
+        Power ratio of largest to second largest mean power strings.
+    """
+
+    chans = sorted(list(wave_bundle.keys()))
+
+    if len(chans) != len(const.rf_channels_ids):
```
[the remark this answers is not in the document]
Mohammad-Neutrino: I thought we decided this to track if the wave_bundle actually returns 16 channels. I can drop it

araproc/analysis/hilbert.py
Mohammad-Neutrino: right, unnecessary!

araproc/analysis/standard_reco.py
Mohammad-Neutrino: ohh, right! sorry for overlooking. will rectify this

araproc/analysis/standard_reco.py
```diff
@@ -819,3 +821,93 @@ def min_frac_corr_depth_multiple(self, *maps, fraction=0.6, z_thresh=0):
...
+                Waveform type requested by which_trace.
+ 
+        Returns
+        -------
+        corr_snr: float
+           Channel pair correlation SNR.
+        """
+
+        corr_func = self.__get_correlation_function(ch1, ch2, wave_packet)
+        corr_snr = hsnr.get_hill_snr(corr_func)
```
marcomuzio, line 854: Good call with this. I think that's the right way to do it since this is hilbert enveloped.

araproc/framework/waveform_utilities.py
```diff
@@ -88,3 +89,34 @@ def freq2time(times, spectrum):
 
     volts = np.fft.irfft(spectrum, axis=-1, n=n_samples_t_domain) * sampling_rate / 2 ** 0.5
     return times, volts
+
+def get_envelope(waveform):
```
marcomuzio, line 93: Can we make this `get_hilbert_envelope` just so it's clear?

araproc/analysis/hilbert.py
```diff
@@ -0,0 +1,73 @@
...
+
+import numpy as np
+from scipy.signal import hilbert
+
+from araproc.analysis import snr
+from araproc.framework import waveform_utilities as wfu
+
+
+
+def get_hill_snr(waveform):
```
marcomuzio, line 11: Can we expand this to `get_hilbert_snr`

araproc/analysis/hilbert.py
```diff
@@ -0,0 +1,73 @@
...
+      return 0
+
+    hill_snr = hill_max/hill_rms
+
+    return hill_snr
+
+
+
+
+def get_avg_hill_snr(wave_bundle, excluded_channels = []):
```
marcomuzio, line 42: and `get_avg_hilbert_snr`

marcomuzio:
@Mohammad-Neutrino thanks this is awesome. I just requested a few tiny changes to name (sorry I'm so nitpicky lol). But this is good to go once those are changed. Great work!

Mohammad-Neutrino:
closing the loop!

PR #22, opened 2024-10-18
Impulsivity related metrices

pgiri4:
I did simple check for RF and CP like events . I find it okay. I have still kept the plotting script within the script hoping that it will help folks to debug this function. The results I saw are here : 
image

image

### DIFF
--- a/araproc/analysis/impulsivity_and_matrices.py
+++ b/araproc/analysis/impulsivity_and_matrices.py
@@ -1,0 +1,113 @@
+import numpy as np
+import matplotlib.pyplot as plt
+from scipy.signal import hilbert
+from scipy.stats import linregress
+from scipy.optimize import curve_fit
+from scipy.special import erf
+
+# Define the erf-linear model for curve fitting
+def erf_linear(x, A, B):
+    return (A * erf(x / B) + x) / (A * erf(1 / B) + 1)
+
+# Generate a Gaussian noise waveform (600 ns)
+np.random.seed(42)
+sampling_interval = 0.5  # ns
+time_array = np.arange(0, 600, sampling_interval)
+voltage_array = np.random.normal(0, 1, len(time_array))
+
+# Function to calculate impulsivity-related variables
+def calculate_impulsivity_measures(voltage_array,time_array):
+    result = {}
+
+    # Hilbert transform to get the envelope of the waveform
+    hilbert_envelope = np.abs(hilbert(voltage_array))
+
+    # Find the index of the maximum value in the Hilbert envelope
+    hill_max_idx = np.argmax(hilbert_envelope)
+    hill_max = hilbert_envelope[hill_max_idx]
+
+    # Sorting based on closeness to the maximum index
+    closeness = np.abs(np.arange(len(voltage_array)) - hill_max_idx)
+    clo_sort_idx = np.argsort(closeness)
+
+    # Sort the Hilbert envelope by closeness to the maximum
+    sorted_waveform = hilbert_envelope[clo_sort_idx]
+
+    # Cumulative distribution function (CDF) calculation
+    cdf = np.cumsum(sorted_waveform)
+    cdf /= np.max(cdf)
+
+    # Linear regression to get slope, intercept, and other statistics
+    slope, intercept, r_value, p_value, std_err = linregress(np.arange(len(cdf)), cdf)
+    cdf_fit = slope * np.arange(len(cdf)) + intercept
+    t_frac = np.linspace(0,1,len(cdf))
+
+    # Calculate Kolmogorov-Smirnov statistic 
+    ks = np.max(np.abs(cdf_fit - cdf))
+    # Perform erf-linear fit on the CDF
+    popt, _ = curve_fit(erf_linear, t_frac, cdf, p0=[intercept / slope, 0.01], bounds=(0, [3 * intercept / slope, 0.5]))
+    A_fit, B_fit = popt
+    cdf_erf_fit = erf_linear(t_frac, A_fit, B_fit)
+
+    # Calculate impulsivity
+    impulsivity = 2 * np.mean(cdf) -1 # np.sum(cdf[1:] * (np.arange(1, len(cdf)) - np.arange(0, len(cdf) - 1))) - 1
+    print(impulsivity)
+    # Calculate linear chi2 (difference between linear fit and ideal x=y line)
+    chi2_linear = np.sum((cdf - np.arange(len(cdf)) / len(cdf)) ** 2)
+
+    # Calculate erf-linear chi2 (difference between erf-linear fit and data)
+    chi2_erf_linear = np.sum((cdf - cdf_erf_fit) ** 2)
+
+    # Store the results
+    result['impulsivity'] = impulsivity
+    result['hill_max'] = hill_max
+    result['slope'] = slope
+    result['intercept'] = intercept
+    result['ks'] = ks  # Kolmogorov-Smirnov statistic
+    result['r_value'] = r_value
+    result['p_value'] = p_value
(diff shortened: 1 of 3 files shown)


araproc/analysis/impulsivity_and_matrices.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # Linear regression to get slope, intercept, and other statistics
+    slope, intercept, r_value, p_value, std_err = linregress(np.arange(len(cdf)), cdf)
+    cdf_fit = slope * np.arange(len(cdf)) + intercept
+    t_frac = np.linspace(0,1,len(cdf))
+
+    # Calculate Kolmogorov-Smirnov statistic 
+    ks = np.max(np.abs(cdf_fit - cdf))
+    # Perform erf-linear fit on the CDF
+    popt, _ = curve_fit(erf_linear, t_frac, cdf, p0=[intercept / slope, 0.01], bounds=(0, [3 * intercept / slope, 0.5]))
```
marcomuzio, line 48: I would place a finite lower bound on parameter B since the function diverges numerically (maybe formally too?) if B is 0. I bounded it to the range (1e-6, 0.5)

araproc/analysis/impulsivity_and_matrices.py
```diff
@@ -0,0 +1,113 @@
...
+    A_fit, B_fit = popt
+    cdf_erf_fit = erf_linear(t_frac, A_fit, B_fit)
+
+    # Calculate impulsivity
+    impulsivity = 2 * np.mean(cdf) -1 # np.sum(cdf[1:] * (np.arange(1, len(cdf)) - np.arange(0, len(cdf) - 1))) - 1
+    print(impulsivity)
+    # Calculate linear chi2 (difference between linear fit and ideal x=y line)
+    chi2_linear = np.sum((cdf - np.arange(len(cdf)) / len(cdf)) ** 2)
```
marcomuzio, line 56: This should be calculated using `cdf_fit` above. Analogous to what you did for the erf version two lines down.

araproc/analysis/impulsivity_and_matrices.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # Calculate erf-linear chi2 (difference between erf-linear fit and data)
+    chi2_erf_linear = np.sum((cdf - cdf_erf_fit) ** 2)
+
+    # Store the results
+    result['impulsivity'] = impulsivity
+    result['hill_max'] = hill_max
```
marcomuzio, line 63: We can remove this, it will be covered by the hilbert snr.

araproc/analysis/impulsivity_and_matrices.py
```diff
@@ -0,0 +1,113 @@
...
+sampling_interval = 0.5  # ns
+time_array = np.arange(0, 600, sampling_interval)
+voltage_array = np.random.normal(0, 1, len(time_array))
+
+# Function to calculate impulsivity-related variables
+def calculate_impulsivity_measures(voltage_array,time_array):
+    result = {}
+
+    # Hilbert transform to get the envelope of the waveform
+    hilbert_envelope = np.abs(hilbert(voltage_array))
```
marcomuzio, line 23: A utility to calculate this will be added to waveform_utilities.py soon, so we'll call that eventually.

araproc/analysis/impulsivity_and_matrices.py
```diff
@@ -0,0 +1,113 @@
...
+    result['slope'] = slope
+    result['intercept'] = intercept
+    result['ks'] = ks  # Kolmogorov-Smirnov statistic
+    result['r_value'] = r_value
+    result['p_value'] = p_value
+    result['std_err'] = std_err
+    result['impLinChi2'] = chi2_linear
+    result['impErfLinChi2'] = chi2_erf_linear
+    result['impErfA'] = A_fit
+    result['impErfB'] = B_fit
```
marcomuzio, line 73: Clever to return a dictionary! I think that's a pretty efficient way to do it.

marcomuzio:
@pgiri4 This is looking really great! I like what you've done here. Can you put the impulsivity stuff into a file `impulsivity.py` without the plotting stuff? Once that is done and the other comments are addressed it should be good. Nice work!

araproc/analysis/impulsivity_and_matrices.py
```diff
@@ -0,0 +1,113 @@
...
+    return (A * erf(x / B) + x) / (A * erf(1 / B) + 1)
+
+# Generate a Gaussian noise waveform (600 ns)
+np.random.seed(42)
+sampling_interval = 0.5  # ns
+time_array = np.arange(0, 600, sampling_interval)
+voltage_array = np.random.normal(0, 1, len(time_array))
+
+# Function to calculate impulsivity-related variables
+def calculate_impulsivity_measures(voltage_array,time_array):
```
marcomuzio, line 19: Lets add our usual documentation about what it does, inputs, and outputs to this function.

araproc/analysis/impulsivity_and_matrices.py
```diff
@@ -0,0 +1,113 @@
...
+from scipy.signal import hilbert
+from scipy.stats import linregress
+from scipy.optimize import curve_fit
+from scipy.special import erf
+
+# Define the erf-linear model for curve fitting
+def erf_linear(x, A, B):
+    return (A * erf(x / B) + x) / (A * erf(1 / B) + 1)
+
+# Generate a Gaussian noise waveform (600 ns)
```
marcomuzio, line 12: We'll also drop this from the final version of course.

araproc/analysis/impulsivity_and_matrices.py
pgiri4: You mean this : chi2_linear = np.sum((cdf_fit - np.arange(len(cdf_fit)) / len(cdf_fit)) ** 2) instead of cdf only , right?

araproc/analysis/impulsivity_and_matrices.py
pgiri4: If I understand you correctly, you are asking for lowering the initial guess of B, doing something like :  popt, _ = curve_fit(erf_linear, t_frac, cdf, p0=[intercept / slope, 1e-6], bounds=(0, [3 * intercept / slope, 0.5])), right?

marcomuzio: No sorry I think the initial guess was fine, I just mean the lower bound shouldn't be 0. So something like 
`popt, _ = curve_fit(erf_linear, t_frac, cdf, p0=[intercept / slope, 1e-2], bounds=([0,1e-6], [3 * intercept / slope, 0.5]))`

araproc/analysis/impulsivity_and_matrices.py
marcomuzio: Haha switched the opposite one! Like this:
`chi2_linear = np.sum((cdf - cdf_fit)**2)`

araproc/analysis/impulsivity_and_matrices.py
```diff
@@ -0,0 +1,125 @@
+import numpy as np
```
marcomuzio, line 1: Let's simplify this file name to `impulsivity.py` to make importing simpler.

araproc/analysis/impulsivity_and_matrices.py
marcomuzio: Please update this to use the `get_hilbert_envelope` function from araproc/framework/waveform_utilities.py.

marcomuzio:
@pgiri4 thanks for making those updates! Just have 2 minor comments left. Once those are addressed this can be merged. Thanks!

marcomuzio:
Perfect, thanks! Feel free to merge!

PR #23, opened 2024-10-18
Add ability to auto load pedestals

clark2668:
Now that we have a common place in cvmfs for pedestal files, we can ask AraProc to load them for us automatically. The user can still override this by passing a pedestal file directly. But passing the pedestal file directly is no longer required.

### DIFF
--- a/araproc/framework/data_visualization.py
+++ b/araproc/framework/data_visualization.py
@@ -14,6 +14,7 @@ https://github.com/matplotlib/matplotlib/issues/20300.
 """
 import matplotlib # noqa : E402
 matplotlib.use("agg")
+__mplver = matplotlib.__version__ # this seems to fix a rare segfault (don't ask me...)
 
 def plot_waveform_bundle(
     waveform_dict = None,
--- a/araproc/framework/data_visualization.py
+++ b/araproc/framework/data_visualization.py
@@ -121,5 +121,6 @@ def plot_skymap(the_map = None,
     ROOT.gPad.SetRightMargin(0.15) # make space for the z axis
     c.SaveAs(ouput_file_path)
     ROOT.gPad.SetRightMargin(0) # reset, so we don't affect settings globally
+    c.Close()
 
     del c
--- a/araproc/framework/dataset.py
+++ b/araproc/framework/dataset.py
@@ -18,13 +18,28 @@ from . import config_files
 def file_is_safe(file_path):
     
     if not isinstance(file_path, str):
-        raise TypeError("Path to file must be a string")
+        raise TypeError(f"Path to file must be a string. Is {type(file_path)}")
     if not os.path.exists(file_path):
         raise FileNotFoundError(f"File ({file_path}) not found")
     if not os.path.isfile(file_path):
         raise ValueError(f"{file_path} looks like a directory, not a file")
     
     return True
+
+def get_cvmfs_ped_file_name(station_id, run_number):
+
+    if station_id not in const.valid_station_ids:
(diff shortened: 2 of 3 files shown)


araproc/framework/data_visualization.py
```diff
@@ -14,6 +14,7 @@
 """
 import matplotlib # noqa : E402
 matplotlib.use("agg")
+__mplver = matplotlib.__version__ # this seems to fix a rare segfault (don't ask me...)
```
marcomuzio, line 17: Enough with your hacks, Clark.

araproc/framework/dataset.py
```diff
@@ -18,14 +18,29 @@
...
         raise ValueError(f"{file_path} looks like a directory, not a file")
     
     return True
 
+def get_cvmfs_ped_file_name(station_id, run_number):
+
+    if station_id not in const.valid_station_ids:
+        raise Exception(f"Station id {station_id} is not supported")
+
+    if not np.isfinite(run_number) and run_number > 0:
```
marcomuzio, line 34: I'm not sure I understand the second condition? Doesn't that allow negative run numbers to move on?

marcomuzio: Unless it's the order of the boolean operations I'm missing?

araproc/framework/dataset.py
```diff
@@ -18,14 +18,29 @@
...
+        raise Exception(f"Station id {run_number} is not supported")
+    
+    cvmfs_top_dir = "/cvmfs/icecube.osgstorage.org/icecube/PUBLIC/groups/arasoft/pedestals"
+    start = start = run_number - (run_number % 1000)
```
marcomuzio, line 38: This seems like a typo?

araproc/framework/dataset.py
clark2668: Nope, that's just a plain typo. Fixed, thanks!

araproc/framework/dataset.py
clark2668: Yup, thank you! Fixed now.

araproc/framework/data_visualization.py
clark2668: Never!

araproc/framework/dataset.py
```diff
@@ -231,7 +248,28 @@ def __assign_config(self):
...
 
-    def __load_pedestal(self):
+    def __load_pedestal(self, path_to_pedestal_file = None):
+
+        # find the right pedestal
+        # if the user provided one, then choose that
+        # otherwise, try to find it in cvmfs
+        # and if that doesn't work, raie an error
+        if path_to_pedestal_file is not None:
+            if file_is_safe(path_to_pedestal_file):
```
marcomuzio, line 258: Do we want a try-except here? I guess `file_is_safe` crashes the code if it isn't safe?

clark2668: I mean, that function has turned into something of a frankenstein... Let me see if I can clean it up a bit. It both returns a true/false, and raises, which isn't great practice really

clark2668: Okay, see what you think now?

clark2668:
Okay, I made requested tweaks, see what you think now.

araproc/framework/dataset.py
marcomuzio: Looks good to me!

marcomuzio:
@clark2668 Looks good to me! Merge if you dare.

PR #24, opened 2024-10-19
Standard reco getter functions for correlation functions

marcomuzio:
This pull request is aimed at helping merge the csw and corr_snr functions into standard_reco.

Cached correlation functions in `StandardReco` object as private variables. Added a private getter function for correlation functions and a public function to look up index of channel pairs.

### DIFF
--- a/araproc/analysis/standard_reco.py
+++ b/araproc/analysis/standard_reco.py
@@ -183,7 +183,7 @@ class StandardReco:
 
         return cross_correlations
 
-    def do_standard_reco(self, waveform_bundle, event_number):
+    def do_standard_reco(self, wavepacket, event_number):
         
         """
         A function to do a standard set of reconstructions.
@@ -211,13 +211,19 @@ class StandardReco:
 
         Parameters
         ----------
-        waveform_bundle : dict
-            A dictionary of the 16 waveforms.
-            The key is the RF channel number.
-            The value is a TGraph.
-            There should be 16 entries, even if you don't intend to use all
-            16 traces in your interferometry.
-            The exclusions are handled further down under the excluded channels section.
+        wavepacket : dict
+            A dict with three entries:
+              "event" : int  
+                Event number
+              "waveforms" : dict
+                A dictionary of the 16 waveforms.
+                The key is the RF channel number.
+                The value is a TGraph.
+                There should be 16 entries, even if you don't intend to use all
+                16 traces in your interferometry.
+                The exclusions are handled further down under the excluded channels section.
+              "type" : string
+                Waveform type requested by which_trace
         event_number : int
             The number of the event corresponding to the traces in wave_bundle. Required
             to access event-specific information like the correlation functions.       
@@ -246,6 +252,9 @@ class StandardReco:
         """
 
         reco_results = {}
+
+        event_number = wavepacket["event"]
+        waveform_bundle = wavepackt["waveforms"]
 
         # update correlation functions if needed
         if(self.__latest_event_num != event_number):
@@ -420,7 +429,7 @@ class StandardReco:
                                                                                        )
         return arrival_time
 
-    def __get_correlation_function(self, ch1, ch2, event_number):
+    def __get_correlation_function(self, ch1, ch2, wavepacket):
         """
         Returns the correlation function for the channel pair ch1-ch2.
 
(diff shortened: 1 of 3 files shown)


araproc/analysis/standard_reco.py
```diff
@@ -389,6 +396,76 @@ def lookup_arrival_time(self,
...
+        Returns
+        -------
+        corr_func : TGraph
+          Cross-correlation function for requested channel pair.
+        """
+
+        if(self.__latest_event_num == -1):
+          raise Exception("No event loaded or event number not set. You must \
+                           run standard_reco.do_standard_reco(wave_bundle, event_num) \
+                           before calling this function. Abort.")
```
clark2668, line 421: Do you insist on this? I can imagine worlds where someone is interested in the corr functions without doing the reconstruction. Why not break off populating `self.__corr_functions_v` and `self.__corr_functions_h` into its own function, which gets called either by (1) `reco.get_standard_reco()` or `reco.__get_correlation_function()`. Makes things slightly more modular. But I won't insist on this. It certainly adds complexity.

examples/run_example.py
```diff
@@ -87,7 +87,9 @@
         print(f"The Average SNR is {avg_snr:.1f}")
  
         # run our standard suite of reconstructions
-        reco_results = reco.do_standard_reco(wave_bundle)
+        reco_results = reco.do_standard_reco(wave_bundle, useful_event.eventNumber)
+
+        pair_idx = reco.get_pair_index(1, 2, reco.pairs_v)
```
clark2668, line 92: should the example also show how to get the correlation function? Just for completeness?

clark2668 (commented):
Okay, asked a "big picture" question, and then made a small suggestion. Otherwise I think this looks pretty good.

examples/run_example.py
marcomuzio: The correlation functions can't be got outside the class because the getter functions are private.

araproc/analysis/standard_reco.py
marcomuzio: Okay ya I will give it a shot. I can see the benefit!

examples/run_example.py
clark2668: Okay right, I forgot. Nevermind!

araproc/analysis/standard_reco.py
marcomuzio: Okay I think I did it. Let me know what you think.

araproc/analysis/standard_reco.py
```diff
@@ -173,14 +211,20 @@ def do_standard_reco(self, waveform_bundle):
...
+              "event" : int  
+                Event number
+              "waveforms" : dict
+                A dictionary of the 16 waveforms.
+                The key is the RF channel number.
+                The value is a TGraph.
+                There should be 16 entries, even if you don't intend to use all
+                16 traces in your interferometry.
+                The exclusions are handled further down under the excluded channels section.
+              "type" : string
```
clark2668, line 225: This is nitpicky, but `type` is a protected work in python. (like `type` means something, e.g. `type(some_variable)`. Can we pick another word?

araproc/analysis/standard_reco.py
```diff
@@ -206,27 +250,21 @@ def do_standard_reco(self, waveform_bundle):
...
-        corr_functions_v = self.rtc_wrapper.correlators["distant"].GetCorrFunctions(
-                                                                    self.pairs_v,
-                                                                    waveform_map
-                                                                    )
-        corr_functions_h = self.rtc_wrapper.correlators["distant"].GetCorrFunctions(
-                                                                    self.pairs_h,
-                                                                    waveform_map
-                                                                    )
+        event_number = wavepacket["event"]
+        waveform_bundle = wavepackt["waveforms"]
```
clark2668, line 254: Typo: `wavepackt`. I'm surprised this even ran?

clark2668 (commented):
Okay, i requested a couple other small changes, but I think this is otherwise good to go.

araproc/analysis/standard_reco.py
marcomuzio: Merp maybe I forgot to rerun it after I introduced that one. It's definitely running fine now.

araproc/analysis/standard_reco.py
marcomuzio: Fixed!

clark2668:
Okay, I'd say let's go for it.

PR #27, opened 2024-10-28
Added significance calculation to impulsivity.

marcomuzio:
Added calculation of significance of erf-linear fit to cdf curve over linear fit using Wilks' theorem.


clark2668:
Seems ok to me?

PR #28, opened 2024-10-28
Updated A1 livetime configurations after full data noise models

Mohammad-Neutrino:


marcomuzio:
Looks good! Merge away.

PR #30, opened 2024-10-28
update excluded channel treatment

clark2668:
This creates a dataset object that contains the excluded channels (according to the yaml file), plus the hpols or the vpols. This way if you want to compute a vpol only variable, you can do like:
```
snr_vpol = snr.get_snr(wavebundle, excluded_channels = d.excluded_channels_and_hpol)
```


marcomuzio:
Looks good to me!

clark2668:
Thank you!

PR #32, opened 2024-10-29
Cache full phased cross-correlation function

marcomuzio:
Added calculation of full-phased (i.e. non-Hilbert enveloped) correlation function. This is cached for later use. Related functions were updated to support this. Added support for making non-Hilbert enveloped reconstruction maps.

### DIFF
--- a/araproc/analysis/standard_reco.py
+++ b/araproc/analysis/standard_reco.py
@@ -168,7 +168,7 @@ class StandardReco:
 
         self.__latest_event_num = -1
 
-    def __calculate_cross_correlations(self, waveform_bundle, pairs):
+    def __calculate_cross_correlations(self, waveform_bundle, pairs, applyHilbert):
 
         """
         Wrapper for cross-correlation calculation.
@@ -183,7 +183,9 @@ class StandardReco:
             16 traces in your interferometry.
             The exclusions are handled further down under the excluded channels section.       
         pairs : std::map<int, std::vector<int> >
-          Channel pairs to form cross-correlations from.
+            Channel pairs to form cross-correlations from.
+        applyHilbert : bool
+            Boolean whether the cross-correlation function is Hilbert enveloped.
  
         Returns
         -------
@@ -198,7 +200,7 @@ class StandardReco:
         for chan_i in waveform_bundle.keys():
             wf_map[chan_i] = waveform_bundle[chan_i]
                 
-        cross_correlations = self.rtc_wrapper.correlators["distant"].GetCorrFunctions(pairs, wf_map)
+        cross_correlations = self.rtc_wrapper.correlators["distant"].GetCorrFunctions(pairs, wf_map, applyHilbert)
         
         del wf_map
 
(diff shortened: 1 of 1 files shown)


araproc/analysis/standard_reco.py
```diff
@@ -452,19 +468,26 @@ def __get_correlation_function(self, ch1, ch2, wavepacket):
...
 
           # get the correlation functions
-          self.__corr_functions_v = self.__calculate_cross_correlations(waveform_bundle, self.pairs_v)
-          self.__corr_functions_h = self.__calculate_cross_correlations(waveform_bundle, self.pairs_h) 
+          self.__corr_functions_v_fullphase = self.__calculate_cross_correlations(waveform_bundle, self.pairs_v, False)
+          self.__corr_functions_h_fullphase = self.__calculate_cross_correlations(waveform_bundle, self.pairs_h, False)
+          self.__corr_functions_v = self.__calculate_cross_correlations(waveform_bundle, self.pairs_v, True)
+          self.__corr_functions_h = self.__calculate_cross_correlations(waveform_bundle, self.pairs_h, True) 
 
         if(ch1//8 != ch2//8):
```
abigailbishop, line 476: Not a part of this PR, but should this check channel polarization directly or do we never expect to put PA data through this framework?

araproc/analysis/standard_reco.py
```diff
@@ -417,7 +431,7 @@ def lookup_arrival_time(self,
                                                                                        )
         return arrival_time
 
-    def __get_correlation_function(self, ch1, ch2, wavepacket):
+    def __get_correlation_function(self, ch1, ch2, wavepacket, applyHilbert):
```
abigailbishop, line 434: Since we're planning on using correlation functions outside of this class, should we drop the double underscore in the naming of this function?

marcomuzio: Are we planning to? I thought the plan was to move csw into standard reco for the final commit?

araproc/analysis/standard_reco.py
marcomuzio: No PA won't ever be put thru AraProc, but it's a good point. There's a better way to do this. I will update.

araproc/analysis/standard_reco.py
abigailbishop, line 434: I have another finicky thought, I'm going to want to use the data from this functions so the smartest way would be to call `__get_correlation_function` then access the items through `__corr_functions_*` right? That way we don't recalculate the cross correlation over and over? 

Super finicky and over the top, sorry feel free to ignore, but would it make sense to call this `__calculate_correlation_function` that only calculates the cross correlations and saves them to `__corr_functions_*` then having a getter function that only calls `__calculate_correlation_function` if those `__corr_functions_*` are not initialized yet (or equal to `None`) and  returns the function the user requested?

abigailbishop: Oh gosh my bad are we? Recinded

araproc/analysis/standard_reco.py
abigailbishop: Cool! There are a couple of other spots that use this notation that I didn't flag.

araproc/analysis/standard_reco.py
marcomuzio: That function actually already exists! It's true this isn't actually _just_ a getter function, but it will only calculate the correlation functions if another event's correlation functions are currently cached. Otherwise it will return the cached correlation function that you request. So you never need to access the `__corr_functions_*` directly. Correct usage would be something like `my_corr = __get_correlation_function(0, 1, wavepacket, False)` Another example is in `get_corr_snr`

marcomuzio: No worries!

abigailbishop: Oh I see, I think I misunderstood it because I was only looking at the diff. Sorry about that. Sounds good to me, thanks!

abigailbishop:
Thanks for doing this Marco and for replying to my questions! Looks good to me :)

araproc/analysis/standard_reco.py
marcomuzio: Oh you're right. I think I got them all now. Thanks!

marcomuzio:
Thanks Abby! Will merge :)

PR #34, opened 2024-10-31
Add Force Trigger Trigger Type for Station 5

abigailbishop:
Add funcitonality to support ARA05's force trigger readout from the Phased Array while performing the DAQ quality cut


marcomuzio:
Thanks Abby! This looks good, I'm going to merge.

PR #35, opened 2024-11-01
Merged configs 3 and 4 & added more CW frequencies

Mohammad-Neutrino:


marcomuzio:
Hi @Mohammad-Neutrino thanks for updating this. Why are so many CW frequencies suddenly being added? Have you found these in data?

Mohammad-Neutrino:
Hi @marcomuzio , so, as you know, Config 1 and 2 are more noisy and had many CWs identified by sigma variance method. After making the average noise models, a few prominent ones showed up particularly in HPol channels. I added those to be on safe side.

marcomuzio:
Ahh, I see okay. Thanks, feel free to merge this!

PR #37, opened 2024-11-01
Update Standard Reco

abigailbishop:
Removed some try/except blocks, save more information to the `reco_results` object, and cleaned some bugs/typos in CSW.

PR #38, opened 2024-11-02
Update Arrival Time wrapper to match RTC Updates

abigailbishop:


marcomuzio:
I think this looks good to merge, but I think some of the commits overlap with PR #37 ? Not sure if they conflict and only one should be merged?

abigailbishop:
Oh whoops yeah. I think if we merge PR#37 first it'll make more logical sense

marcomuzio:
Okay seems that's right! I'm gonna merge this.

PR #41, opened 2024-11-04
CSW Plotting Function

abigailbishop:
Adds a CSW plotting function to standard_reco.py. Added an option to get_csw to return the final waveforms for each channel as they were after being modified and right before being added to the CSW. These CSW components are required by the plotting script.


marcomuzio:
Hi @abigailbishop, thanks for putting this together seems like it's almost ready! I have a few thoughts. 

I think that a script that takes the run and event number as arguments would be helpful and should probably live in `FiveStation/scripts/L2`? Any helper plotting function probably belongs in `framework/data_visualization.py`. 

My main concern is whether this requires us to perform a full reconstruction to use? One could imagine reading in the L2 results for an event, getting its theta/phi/distance for the max corr and the handing that to a function which plots the corresponding CSW. That might be useful to better reproduce the L2 results, instead of relying on the reco to return the same results (which maybe it wouldn't if we didn't reco over the same maps again).

abigailbishop:
Thanks Marco and sorry for the delay in getting back to this. 

I moved the plotting function to `framework/data_visualization.py` and I changed the CSW calculation function to take the `theta` and `phi` values needed for the arrival delays calculated via reconstruction directly as function parameters instead of passing `reco_results` and accessing the theta and phi through that object. 

I can work on the script that takes a run and event number and returns the CSW, though it'll take some time to write something that'll find the waveforms for an event given our `/data/exp/ARA/{year}/L1/{pct}pct/{station}/{day}/run{run}/event{run}.root` organization scheme. If we already have something like this in AraProc or FiveStation, that would make this much easier. We also need to have a way to map entry numbers to event numbers in the 10pct samples.

abigailbishop:
Also, which file do you suggest this CSW plotter helper function should live in? `FiveStation/fivestation/visualize.py`?

marcomuzio:
Hi Abby, thanks for making these changes! Will take a closer look after I land today. For now, I think that `visualize.py` seems a sensible place to put the helper function. For some inspiration on how to find the files, take a look at [here](https://github.com/ara-software/FiveStation/blob/380c188e4f5585bfb659f6949f731649202787f8/scripts/L2/getEventSummary.py#L24). `glob` is your friend, just need to wildcard over the directories you want to search basically. Also, to the extent possible we should only work in terms of the `event_number`. That is also done in the same script [here](https://github.com/ara-software/FiveStation/blob/380c188e4f5585bfb659f6949f731649202787f8/scripts/L2/getEventSummary.py#L43).

marcomuzio:
Hey Abby, sorry did you commit the changes you mentioned above? I don't see them here.

abigailbishop:
No sorry I didn't push them to GitHub. Just pushed but now I need to move the csw plotter to the other repo

abigailbishop:
Okay I think that's all the updates needed to AraProc so, unless you have more suggestions, I'm done with this PR and will switch to a CSW PR for FiveStation

marcomuzio:
Nope, all looks good to me! I think we can merge.

PR #44, opened 2024-11-04
Fixed min depth corr calculation.

marcomuzio:
Accounted for the fact that the max correlation can be above the ice surface by removing the `z_thresh` condition. Fixed the calculation across maps to use the maximum correlation value across all maps as the reference value. Fixed the A1 calpulser radius.

### DIFF
--- a/araproc/analysis/standard_reco.py
+++ b/araproc/analysis/standard_reco.py
@@ -96,7 +96,7 @@ class StandardReco:
         # so look that up
 
         self.calpulser_r_library = {
-            100 : "49.48",
+            100 : "48.02",
             2 : "42.86",
             3 : "41.08",
             4 : "52.60",
@@ -754,11 +754,10 @@ class StandardReco:
         
         return surf_corr_ratio, max_surf_corr_result, max_corr_result
 
-    def min_frac_corr_depth(self, corr_map, fraction=0.6, z_thresh=0):
+    def min_frac_corr_depth(self, corr_map, max_corr, fraction=0.6):
         """
         Finds the shallowest depth where correlation meets a specified fraction of the maximum correlation,
         using 'theta' to calculate depth and adjusting by the average z-coordinate of antennas.
-        Ensures the result remains below the specified depth threshold (z_thresh).
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


araproc/analysis/standard_reco.py
```diff
@@ -96,7 +96,7 @@ def __init__(self,
         # so look that up
 
         self.calpulser_r_library = {
-            100 : "49.48",
+            100 : "48.02",
```
clark2668, line 99: Is this intended? It keeps changing....

araproc/analysis/standard_reco.py
```diff
@@ -803,16 +801,13 @@ def min_frac_corr_depth(self, corr_map, fraction=0.6, z_thresh=0):
                     depth = radius * math.sin(math.radians(theta)) + avg_z
                     
 
-                    # Ensure min_depth does not exceed z_thresh
-                    if depth > min_depth and depth <= z_thresh:
+                    # Update min_depth 
+                    if depth > min_depth:
```
clark2668, line 805: This goes the right way? In case depth is a negative number?

AlanSalcedo: Right. I remember this is sweeping downward from the surface (really from z_threshold). Maybe we need to sweep up for negative depths? 

I can do that today if you want

clark2668: I don't have strong opinions about which way you go! I'm really asking about the sign convention I guess.

araproc/analysis/standard_reco.py
marcomuzio: Yes, this is restoring the value that was changed in [PR#42](https://github.com/ara-software/AraProc/pull/42). I did this based on the values Alex said to ignore/take seriously.

araproc/analysis/standard_reco.py
marcomuzio: This should be right I think. A more negative depth is closer to Earth's center and you want the shallowest (i.e. most positive) depth where the conditions are satisfied.

marcomuzio:
All happy with this?

araproc/analysis/standard_reco.py
clark2668: Okay, then sounds good

araproc/analysis/standard_reco.py
clark2668: Okay, great

clark2668:
Merging sounds good to me!

AlanSalcedo:
Looks good to me. Thanks Marco!

marcomuzio:
Thanks guys!

PR #45, opened 2024-11-05
Known bad runs and bad events by unix time

AlanSalcedo:
Bad runs are identified from lists containing known bad runs. Here's where I got them from:

https://github.com/abigailbishop/MF_filters/blob/main/tools/ara_known_issue.py#L27

All events in a bad run are marked as bad in one go. 

Individual bad events are identified by their unixTimes. I added the ones for A2/3 also from:

https://github.com/abigailbishop/MF_filters/blob/main/tools/ara_known_issue.py#L27

These are identified in an event-by-event basis.

Main changes live in: ara_known_issues.py and bad_run_quality_cut.py. All other files are lists with known bad runs.

Note the comment in L443 of ara_known_issues.py:

        Year = np.arange(2014, 2020, dtype = int) #ASG: What years?

I need to create known bad run files for years 2020-2023.

Also note hardcoded paths: 

'/home/alansalgo/ARA/AraProc/araproc/analysis/data/Data_processing_and_storage_plan/'

Do we want to create a $dataPath later?


AlanSalcedo:
Also do we want the quality cuts in AraProc or FiveStation?

marcomuzio:
Sorry I forgot to comment here. Both of these cuts should go into the cuts.py module of FiveStation. The bad run cut should pull from the run_logs directory. So the user may need to point the Cuts class to where those are.

marcomuzio:
But also, maybe I don't understand the bad unix times thing exactly either. I naively thought this was to remove specific periods of time from the analysis not individual events?

PR #48, opened 2024-11-08
make finding the surface correlation value faster

clark2668:
This uses the ROOT internal functions to find the surface correlation value, which is substantially faster than the manual for loops in python. This runs in ~3.5E-5 seconds/map, where manually looping takes 0.17 seconds/map. For loops in python are evil! :)

### DIFF
--- a/araproc/analysis/standard_reco.py
+++ b/araproc/analysis/standard_reco.py
@@ -582,34 +582,22 @@ class StandardReco:
         if hist is None:
             raise ValueError("The 'map' key was not found in corr_map.")
         
-        surf_corr_values = []
-        corr_bins = []
-
-        # Loop through the bins and store values within theta range
-        for x_bin in range(1, hist.GetNbinsX() + 1):
-            for y_bin in range(1, hist.GetNbinsY() + 1):
-                theta = hist.GetYaxis().GetBinCenter(y_bin)
-                phi = hist.GetXaxis().GetBinCenter(x_bin)
-                corr = hist.GetBinContent(x_bin, y_bin)
-                if theta_min <= theta <= theta_max:
-                    surf_corr_values.append(corr)
-                    corr_bins.append((x_bin, y_bin))
-
-        # Identify maximum surface correlation
-        if not surf_corr_values:
-            raise RuntimeError("No correlation values found in the specified theta range.")
+        # set the range of the histogram to only be where it's relevant
+        theta_min_bin = hist.GetYaxis().FindBin(theta_min)
+        theta_max_bin = hist.GetYaxis().FindBin(theta_max)
+        hist.GetYaxis().SetRange(theta_min_bin+1, theta_max_bin-1)
         
-        # Check for NaNs
-        if np.isnan(surf_corr_values).any():
-            raise ValueError("surf_corr_values contains NaN values.")
+        # locate the peak in that range, and return the location
+        _peakZ = ctypes.c_int()
+        _peakTheta = ctypes.c_int()
+        _peakPhi = ctypes.c_int()
+        hist.GetMaximumBin(_peakPhi, _peakTheta, _peakZ)
+        max_phi = hist.GetXaxis().GetBinCenter(_peakPhi.value)
+        max_theta = hist.GetYaxis().GetBinCenter(_peakTheta.value)
+        max_surf_corr = hist.GetMaximum()
 
-        # Calculate maximum and index
-        max_surf_corr = np.max(surf_corr_values)
-        max_idx = np.argmax(surf_corr_values)
-        max_bin = corr_bins[max_idx]
-        
-        max_theta = hist.GetYaxis().GetBinCenter(max_bin[1])
-        max_phi = hist.GetXaxis().GetBinCenter(max_bin[0])
+        # reset the range
+        hist.GetYaxis().SetRange(0,0)
 
         return max_surf_corr, max_theta, max_phi
 


araproc/analysis/standard_reco.py
```diff
@@ -582,34 +583,22 @@ def get_surface_corr_max(self, corr_map, z_thresh=-10):
...
-                    surf_corr_values.append(corr)
-                    corr_bins.append((x_bin, y_bin))
-
-        # Identify maximum surface correlation
-        if not surf_corr_values:
-            raise RuntimeError("No correlation values found in the specified theta range.")
+        # set the range of the histogram to only be where it's relevant
+        theta_min_bin = hist.GetYaxis().FindBin(theta_min)
+        theta_max_bin = hist.GetYaxis().FindBin(theta_max)
+        hist.GetYaxis().SetRange(theta_min_bin+1, theta_max_bin-1)
```
marcomuzio, line 589: Why the +1 and -1? I wonder if this would be more straight forward with `hist.GetYaxis().SetRangeUser(theta_min, theta_max)`? But maybe the approximations it makes are a problem.

araproc/analysis/standard_reco.py
```diff
@@ -791,20 +780,15 @@ def min_frac_corr_depth(self, corr_map, max_corr, fraction=0.6):
 
         min_depth = -float('inf')  # Initialize to find the shallowest depth (least negative)
```
marcomuzio, line 781: Can we move this to an `else` below now? I think it'll just be a little clearer that way now.

araproc/analysis/standard_reco.py
```diff
@@ -5,6 +5,7 @@
 import ROOT
 from scipy.interpolate import Akima1DInterpolator
 import itertools
+import time
```
marcomuzio, line 8: Maybe we should drop this now that it's not being used?

araproc/analysis/standard_reco.py
marcomuzio: Neat trick btw!

marcomuzio:
Thanks for making these changes Brian! I left a few small comments, but I think it looks good to merge.

araproc/analysis/standard_reco.py
```diff
@@ -582,34 +583,22 @@ def get_surface_corr_max(self, corr_map, z_thresh=-10):
...
-        # Identify maximum surface correlation
-        if not surf_corr_values:
-            raise RuntimeError("No correlation values found in the specified theta range.")
+        # set the range of the histogram to only be where it's relevant
+        theta_min_bin = hist.GetYaxis().FindBin(theta_min)
+        theta_max_bin = hist.GetYaxis().FindBin(theta_max)
+        hist.GetYaxis().SetRange(theta_min_bin+1, theta_max_bin-1)
         
-        # Check for NaNs
-        if np.isnan(surf_corr_values).any():
```
marcomuzio, line 603: I actually have been getting this error sometimes. Do you know why? Is it safe to remove this?

araproc/analysis/standard_reco.py
clark2668: I think it's the approximations that are biting me here. Somehow using the SetRange functions (either in bin space or user space) doesn't precisely replicate the for loop, and I'm not sure why. The +1 and -1 was the only way I got this to replicate the for loop exactly.

clark2668: I'm not sure what to do about it. Any thoughts?

araproc/analysis/standard_reco.py
clark2668: Yup, will do

araproc/analysis/standard_reco.py
clark2668: I have absolutely no idea why this would happen...

marcomuzio: Ya this is weird, it looks like it would only happen if the map itself had nans in it. I guess that's not the RTC behavior?

clark2668: That's right. The map should never have nan's in it...

clark2668: If it does, I'd like to know so I can address it. Can you identify a specific event where this is firing?

marcomuzio: It fires on some bad quality events, but it also fired on event index 5 of run 5406 on A5 which is apparently a good quality event.

marcomuzio: This is the path if you want to check it out `/data/exp/ARA/2019/blinded/L1/ARA05/0605/run005406/event005406.root`

marcomuzio: Nevermind, I can't reproduce this. Maybe AraRoot was being recompiled or something and it got borked.

araproc/analysis/standard_reco.py
```diff
@@ -572,44 +572,30 @@ def get_surface_corr_max(self, corr_map, z_thresh=-10):
...
-                corr = hist.GetBinContent(x_bin, y_bin)
-                if theta_min <= theta <= theta_max:
-                    surf_corr_values.append(corr)
-                    corr_bins.append((x_bin, y_bin))
-
-        # Identify maximum surface correlation
-        if not surf_corr_values:
-            raise RuntimeError("No correlation values found in the specified theta range.")
+        # set the range of the histogram to only be where it's relevant
+        binwidth = (hist.GetYaxis().fXmax - hist.GetYaxis().fXmin) / float(hist.GetYaxis().fNbins);
```
clark2668, line 584: Sorry, why not `hist.GetYaxis().GetBinWidth(<some_bin>)`?

clark2668 (commented):
Made a small change request

araproc/analysis/standard_reco.py
clark2668: I think you did this.

marcomuzio:
Is this set?

PR #49, opened 2024-11-08
Fix Occasional Sign Error in CSW Fine Tuning 

abigailbishop:
For the CSW, for channels with a channel number greater than the reference channel, you need to multiply the time array of the cross correlation results by -1 so you're comparing the current channel's signal timing to the reference channel's timing, not vice versa. This is because AraRoot always calculated cross correlations by comparing lower numbered channels with respect to higher numbered channels. 

I accounted for this already when saving the fine-tuned arrival delay, but I neglected to consider this while using the reconstructed arrival time delays to fine tune the waveform shifting via cross correlation. So any CSW with a reference channel less than the number of total channels had a slightly odd CSW as a result. 

I ended up reformatting the handling of this to account for the sign flip right after the cross correlation is read in to reduce confusion about the two locations requiring a sign flip.

### DIFF
--- a/araproc/analysis/standard_reco.py
+++ b/araproc/analysis/standard_reco.py
@@ -1060,10 +1060,15 @@ class StandardReco:
                     )
                 )
 
+                # AraRoot always compares channels with smaller IDs to channels with 
+                #   larger IDs but we always want to compare to the reference channel.
+                # Correct for this if the current ch_ID is larger than the reference ch_ID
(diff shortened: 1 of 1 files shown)


marcomuzio:
Looks good to me! Thanks Abby!

clark2668:
Agreed. Rock and roll.

abigailbishop:
Fabulous thank you!

PR #50, opened 2024-11-08
Fixing correlation SNR calculation

Mohammad-Neutrino:
Calling full phase correlation was returning correlation SNR with values `>10**15` due to near zero values in leading and trailing samples. Fixed it by removing those. Here is an example of trimmed (now) vs previous correlation function.
corr_func

### DIFF
--- a/araproc/analysis/standard_reco.py
+++ b/araproc/analysis/standard_reco.py
@@ -865,9 +865,9 @@ class StandardReco:
         corr_snr: float
            Channel pair correlation SNR.
         """
-
-        corr_func = self.__get_correlation_function(ch1, ch2, wave_packet, False)
-        corr_snr = snr.get_snr(corr_func)
+        
+        _, corr_func = wfu.tgraph_to_arrays(self.__get_correlation_function(ch1, ch2, wave_packet, False))
+        corr_snr = snr.get_snr(corr_func[abs(corr_func) > 0.001])
 
         return corr_snr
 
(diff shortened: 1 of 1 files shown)


araproc/analysis/standard_reco.py
```diff
@@ -865,9 +865,9 @@ def get_corr_snr(self, ch1, ch2, wave_packet):
         corr_snr: float
            Channel pair correlation SNR.
         """
-
-        corr_func = self.__get_correlation_function(ch1, ch2, wave_packet, False)
-        corr_snr = snr.get_snr(corr_func)
+        
+        _, corr_func = wfu.tgraph_to_arrays(self.__get_correlation_function(ch1, ch2, wave_packet, False))
+        corr_snr = snr.get_snr(corr_func[abs(corr_func) > 0.001])
```
marcomuzio, line 870: One thing that worries me about this is that it has the effect of cutting points out around zero even in the middle of the correlation function where we don't think there's a problem. So maybe this could introduce an artifact or inflate the rms slightly. Instead, I would remove all points from the beginning of the waveform until the first point with absolute value > 1e-6  (just being generous with the threshold) and then all from the end with the same condition. 

Something like this would do it:
```
 _, corr_func = wfu.tgraph_to_arrays(self.__get_correlation_function(ch1, ch2, wave_packet, False))
 
# trim correlation function 
corr_thresh = 1e-6
idxAboveThresh = np.squeeze(np.where(np.abs(corr_func) >= corr_thresh)) # all indices above threshold
idxFirst = idxAboveThresh[0] # first above threshold
idxLast = idxAboveThresh[-1] # last above threshold
corr_func = corr_func[idxFirst:idxLast+1] # trim to above threshold region
 
# calculate snr
corr_snr = snr.get_snr(corr_func)

Mohammad-Neutrino: Ohh right! I didn't think it through. Making these changes. Thanks, Marco!

marcomuzio:
Looks great! Merge away!

PR #54, opened 2024-11-15
fix broken examples, it was calling function get_waveforms which was  renamed to get_wavepacket in dataset.py file

pgiri4:
Here is a quick fix in existing examples .


marcomuzio:
Looks good! Thanks for fixing these. Merge away!

PR #55, opened 2024-11-16
Fixed impulsivity significance calculation.

marcomuzio:
Fixed impulsivity significance calculation. Both that and the linear chi2 are now correctly based on the null hypothesis of a fixed linear function `cdf=t_frac`  rather than the fit linear function `cdf = m*t_frac + b`. Made some other small improvements to the impulsivity module.


clark2668:
I'm not qualified to evaluate this, but if you've tested this and are happy with it, merge.

marcomuzio:
Thanks!

PR #56, opened 2024-11-17
Add dda voltage cut for all stations per config

pgiri4:
This is the first draft of the DDA voltage cut. A few things: 

1). I have used uproot, but I need suggestions on implementing a PyROOT version of this. I am having problems reading the branch ddaVoltageCurrent[4] correctly with PyROOT. 

2). Will we have a module that reads housekeeping data (event/sensor) and provides us with information per event? 

Please help me to make this better with suggestions and comments.

### DIFF
--- a/araproc/analysis/daq_quality_cut.py
+++ b/araproc/analysis/daq_quality_cut.py
@@ -1,6 +1,10 @@
 import numpy as np
 import logging
 import yaml
+import uproot
+import os
+import re
+
 from araproc.framework import constants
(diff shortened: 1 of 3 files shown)


marcomuzio:
Hey Pawan! After your and @Mohammad-Neutrino 's investigations into this, I think we should move this to be part of the bad run identification script in FiveStation. Given that you all have shown that this often affects a small number of events in a run, let's actually have the function identify a span of bad unix time to be cut, rather than the entire run. Let me know if you have questions!

PR #59, opened 2024-11-19
Undo temporary timing table path fix

marcomuzio:
Updated timing table paths to their permanent locations. Also update `run_example.py` to fix typo that was fixed in [PR#58](https://github.com/ara-software/AraProc/pull/58).


clark2668:
Rock and roll

marcomuzio:
Thanks!

PR #60, opened 2024-11-20
Fast event idx lookup, dynamic spectrum range

marcomuzio:
Added fast event index lookup for a requested event number. This is more efficient than a for loop to find the ROOT file index corresponding to a given event number. Also made spectrum plots y range dynamic.

Finally, added some functionality for if we ever want to check the station id that's stored in the ROOT data (stored as a class variable `data_station_id`. In the ended I decided not to use it, but I already coded this up and maybe it will be useful later. Can remove if wanted.

### DIFF
--- a/araproc/framework/data_visualization.py
+++ b/araproc/framework/data_visualization.py
@@ -70,6 +70,7 @@ def plot_waveform_bundle(
                               layout="constrained")
 
     # draw the graphs, label each one appropriately
+    ymin = 1e100
     ymax = -1e100
     for wave_key in tgraphs_to_plot.keys():
         times, volts = wu.tgraph_to_arrays(tgraphs_to_plot[wave_key])
@@ -83,6 +84,7 @@ def plot_waveform_bundle(
             xvals = freqs*1e3 # from GHz to MHz
             yvals = 10*np.log10(np.abs(spectrum)**2 / 50 / 1e3) # from mV to dBm
 
+        ymin = min(ymin, yvals.min())
         ymax = max(ymax, yvals.max())
 
         axd[f"ch{wave_key}"].plot(xvals, yvals) # make the plot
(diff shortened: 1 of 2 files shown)


araproc/framework/data_visualization.py
```diff
@@ -81,6 +82,8 @@ def plot_waveform_bundle(
             xvals = freqs
             yvals = np.log10(np.abs(spectrum))
```
Mohammad-Neutrino, line 83: Thanks Marco for making things more dynamic. Now that you have opened this PR, could you kindly make the power axis in terms of **dBm** and frequency in **MHz**? I think 
```suggestion
            xvals = freqs*1e3 # from GHz to MHz
            yvals = 10*np.log10(np.abs(spectrum)**2 / 50 / 1e3) # from mV to dBm
```
will do the job.

Mohammad-Neutrino: Accordingly, we can change 
line 44 to
```suggestion
           "freq" : "Frequency (MHz)"
```
line 48 to 
```suggestion
           "freq" : "Spectrum (dBm)"
```

araproc/framework/data_visualization.py
```diff
@@ -92,7 +95,7 @@ def plot_waveform_bundle(
     
     # make limits look nice
     if time_or_freq == "freq":
-        ax.set_ylim([1,4])
+        ax.set_ylim([1,ymax+0.2])
```
Mohammad-Neutrino, line 98: Also ylim of `ax.set_ylim(-50, 50)` should catch the whole y axis in `dBm` but you can go with `-50 to ymax+1` or something similar for `ymin` as `ymin-1`

araproc/framework/dataset.py
```diff
@@ -144,6 +144,7 @@ def __init__(self,
         self.event_tree = None
         self.run_number = None
         self.station_id = None
+        self.data_station_id = None
```
Mohammad-Neutrino, line 147: I like the idea of keeping the `stationID`. Thanks for adding that!

marcomuzio:
Just fyi @Mohammad-Neutrino decided to switch the spectrum frequency axis back to GHz since it's the natural unit for AraProc and also our filters and stuff are in GHz. So hopefully will minimize confusion.

PR #62, opened 2024-11-21
Trigger Type Typo for A5?

AlanSalcedo:


marcomuzio:
Looks good. Approved. Thanks for catching that.

PR #64, opened 2024-11-21
CSW Warning Updates

abigailbishop:
Add comments to areas where the CSW's warning tracker is incremented. Reworked the CSW warning values as I realized I planned them poorly (not incorrectly, but not intuitively) before.

### DIFF
--- a/araproc/analysis/standard_reco.py
+++ b/araproc/analysis/standard_reco.py
@@ -1239,6 +1239,14 @@ class StandardReco:
                 #   waveform 2 yeilds (1.6 - 1.4) % 0.5 = 0.2 
                 #   and waveform 3 yeilds (0.9 - 0.5) % 0.5 = 0.4 
             ) 
+
+            # It was noticed that after subtracting arrival_delays[ch_ID] 
+            #   from the time array, usually all bins start and end on the same 
+            #   fraction of a nanosecond, meaning there should be no need for a
+            #   rebinning shift. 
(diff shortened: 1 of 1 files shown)


araproc/analysis/standard_reco.py
```diff
@@ -1287,22 +1295,24 @@ def get_csw(
...
                 if times[0] <= expected_signal_time <= times[roll_shift_bins]: 
-                    warning += 10_00
+                    warning += 1_00_00_00
             elif roll_shift_bins < 0 and abs(roll_shift_bins)<len(times): 
-                # Rolling from back to front, check that signal region isn't in the back
+                # Rolling from back to front, warn if the expected signal is
+                #   in the region about to be rolled
                 if  times[roll_shift_bins]  <= expected_signal_time <= times[-1]: 
-                    warning += 10_00
+                    warning += 1_00_00_00
```
marcomuzio, line 1315: is this supposed to be different from the one above? (idk if I'm following the logic of these warning, but the rest are different from each other)

marcomuzio:
Just one question, but other than that I think we can merge.

araproc/analysis/standard_reco.py
abigailbishop: I figured these are similar enough (since they both involve rolling the region with expected signal) but they can have different numbers too

marcomuzio: Fine with me. Just wanted to make sure it was intended. I'm happy either way.

PR #66, opened 2024-11-25
Add CW filters from outlier invesitgations

abigailbishop:


araproc/framework/config_files/analysis_configs.yaml
```diff
@@ -86,7 +86,9 @@ station2:
       filt1 : { min_freq : 0.124, max_freq : 0.126, min_power_ratio : 0.02 }
       filt2 : { min_freq : 0.149, max_freq : 0.151, min_power_ratio : 0.02 }
       filt3 : { min_freq : 0.249, max_freq : 0.251, min_power_ratio : 0.02 }
+      filt5 : { min_freq : 0.400, max_freq : 0.500, min_power_ratio : 0.10 }
```
marcomuzio, line 89: Looks like this key is repeated?

clark2668: Also, if you put in this generic 100 MHz wide filter, does that make the two narrow filters after it redundant?

abigailbishop: I agree with Marco about the minimum power ratios. For the order, do you think maybe smaller-frequency-range threshold filters should come first so they get handled before the wider filter? I'm struggling to imagine the advantage of that

marcomuzio: I'm not sure... originally I was thinking ordering so narrower filters come first made sense so that the wide one acts as a sort of "catch-all, cleanup" after? But in practice idk if that makes sense. Now I'm wondering if we should actually order by decreasing min power threshold. In case large peaks aren't fully subtracted, the later filters would have a second chance to remove the remaining peak. 🤷‍♂️🤷‍♂️

abigailbishop: I like the decreasing argument for now until we have a chance to investigate further

marcomuzio:
Not sure if you are waiting on anything else for this, but all good from my end. Merge if you are ready!

clark2668:
Agreed, let's roll

PR #70, opened 2024-12-04
Add waveform quality check.

marcomuzio:
Added a module to perform quality checks on the waveform itself. To avoid errors during processing this needs to be separate from the daq quality check. Currently this just checks for waveform saturation, but wrote it so other checks can be easily added later.

### DIFF
--- a/araproc/analysis/waveform_quality_cut.py
+++ b/araproc/analysis/waveform_quality_cut.py
@@ -1,0 +1,86 @@
+import numpy as np
+import ROOT
+
+from araproc.framework import waveform_utilities as wfu
+
+def check_waveform_saturation(wave_bundle, excluded_channels=[]):
+
+    """
+    Checks whether any channel's trace is saturated.
+
+    Parameters
+    ----------
+    wave_bundle: dict of TGraphs or np.ndarrays
+        Dictionary of waveform TGraphs or np.ndarrays to be averaged.
+    excluded_channels: list
+        List of dictionary keys to exclude from average.
+
+    Returns
+    -------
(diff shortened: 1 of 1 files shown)


drparamita (commented):
Looks fine.

araproc/analysis/waveform_quality_cut.py
```diff
@@ -0,0 +1,86 @@
...
+    -------
+    is_saturated : bool
+        Whether a waveform is saturated.
+    """
+    
+    chans = list(wave_bundle.keys())
+    
+    # set parameters for saturation check
+    n_points = 5
+    saturation_threshold = 1500 # mV
```
clark2668, line 28: Did you confirm that this fires? I wonder if we want to be a little more generous, like 1495 or something?

clark2668: Okay, great, I'm glad it grabbed the sample. Event. Can you add a comment to the code about which one? And okay, fair enough re: 1500

marcomuzio: So you want something like 
`# Example saturation event: A4 run 6421 event 15333`?

clark2668: Exactly

marcomuzio: Okay done in [https://github.com/ara-software/AraProc/pull/71](PR#71)

clark2668: Great, thank you.

PR #75, opened 2024-12-19
First version of the addition of landmarks in AraProc

pgiri4:
Comments are welcome!
A4
image
A1
image

### DIFF
--- a/araproc/analysis/standard_reco.py
+++ b/araproc/analysis/standard_reco.py
@@ -579,7 +579,7 @@ class StandardReco:
         radius = float(radius)
 
         # Calculate average antenna z coordinates
-        _, _, avg_z = mu.calculate_avg_antenna_xyz(self.station_id, self.num_channels)
+        _, _, avg_z = mu.initialize_ARA_geometry(self.station_id).get_station_center(excluded_channels = [])
 
         # Check if surface is visible at the given radius
         if radius < (abs(avg_z) + z_thresh):
@@ -786,7 +786,7 @@ class StandardReco:
         radius = float(radius)
 
         # Calculate average antenna z-coordinate
-        _, _, avg_z = mu.calculate_avg_antenna_xyz(self.station_id, self.num_channels)
+        _, _, avg_z = mu.initialize_ARA_geometry(self.station_id).get_station_center(excluded_channels = [])
 
         # Find the shallowest bin above threshold.
         # We want the last bin ABOVE because theta goes from -90 to 90,
--- a/araproc/analysis/standard_reco.py
+++ b/araproc/analysis/standard_reco.py
@@ -579,7 +579,7 @@ class StandardReco:
         radius = float(radius)
 
         # Calculate average antenna z coordinates
-        _, _, avg_z = mu.initialize_ARA_geometry(self.station_id).get_station_center(excluded_channels = [])
+        _, _, avg_z = mu.AraGeom(self.station_id).get_station_center(excluded_channels = [])
 
         # Check if surface is visible at the given radius
         if radius < (abs(avg_z) + z_thresh):
@@ -786,7 +786,7 @@ class StandardReco:
         radius = float(radius)
 
         # Calculate average antenna z-coordinate
-        _, _, avg_z = mu.initialize_ARA_geometry(self.station_id).get_station_center(excluded_channels = [])
+        _, _, avg_z = mu.AraGeom(self.station_id).get_station_center(excluded_channels = [])
 
         # Find the shallowest bin above threshold.
         # We want the last bin ABOVE because theta goes from -90 to 90,
--- a/araproc/framework/constants.py
+++ b/araproc/framework/constants.py
@@ -15,5 +15,4 @@ num_samples_per_block = 64
 num_samples_per_dda = 32768
 num_polarization = 2
 num_trigger_type = 3
-num_RF_channels = 16
-
+num_rf_channels = len(rf_channels_ids)
--- a/araproc/framework/constants.py
+++ b/araproc/framework/constants.py
@@ -15,3 +15,5 @@ num_samples_per_block = 64
 num_samples_per_dda = 32768
 num_polarization = 2
 num_trigger_type = 3
+num_RF_channels = 16
+
--- a/araproc/framework/data_visualization.py
+++ b/araproc/framework/data_visualization.py
@@ -117,7 +117,7 @@ def plot_waveform_bundle(
     plt.close(fig)
     del fig, axd
 
-def plot_skymap(the_map = None,
+def plot_skymap(st,list_of_landmarks = None,cal_pulse_index = None,spice_depth = None,the_map = None,
                 output_file_path = None
                 ):
     
@@ -126,7 +126,7 @@ def plot_skymap(the_map = None,
 
     if not isinstance(output_file_path, str):
         raise TypeError("Path to output file must be a string")
-
+   
     corr_peak, peak_phi, peak_theta = mu.get_corr_map_peak(the_map)
     the_map.SetTitle(f"Peak Phi/Theta/Corr = {peak_phi:.1f}/ {peak_theta:.1f}/ {corr_peak:.2f}")
     the_map.GetXaxis().SetTitle("Phi (deg)")
@@ -140,7 +140,38 @@ def plot_skymap(the_map = None,
 
     c = ROOT.TCanvas("c", "c", 700, 500)
     c.cd()
-    the_map.Draw("z aitoff")
+    the_map.Draw("COLZ") # keeping this off for now: the_map.Draw("z aitoff")
+
+    ## Add known locations to the skymap 
+    landmark_dict = mu.get_known_landmarks(st,list_of_landmarks,cal_pulse_index,spice_depth)
+    markers = []  # Keep references to markers to avoid garbage collection
+    labels = []   # Keep references to labels
+
+    for entry in landmark_dict.keys():
+        phi = landmark_dict[entry][2]
+        theta = landmark_dict[entry][1]
+
+        # Draw the marker
+        marker = ROOT.TMarker(phi, theta, 29)  # Style 29: Star
+        marker.SetMarkerColor(ROOT.kBlack if entry == "CP" else ROOT.kRed)  # Black for CP, Red for others
+        marker.SetMarkerSize(2.0)
+        marker.Draw("SAME")
+        markers.append(marker)
+
+        # Draw the label
+        label = ROOT.TLatex(phi + 2, theta - 2, entry)  # Offset for clarity
+        label.SetTextColor(ROOT.kWhite)
+        label.SetTextSize(0.03)
+        label.Draw("SAME")
+        labels.append(label)
+
+        if entry == "ICL":
+           vertical_line = ROOT.TLine(phi, -90, phi, 90)  # Draw line from theta=-90 to theta=90
+           vertical_line.SetLineColor(ROOT.kBlue)
+           vertical_line.SetLineStyle(2)  # Dashed line
+           vertical_line.SetLineWidth(2)
+           vertical_line.Draw("SAME")
+
     ROOT.gStyle.SetPalette(112) # viridis
     ROOT.gPad.SetRightMargin(0.15) # make space for the z axis
     c.SaveAs(output_file_path)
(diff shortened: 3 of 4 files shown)


marcomuzio:
Starting to take a look at this! Could you add an example of the output to the description above?

araproc/framework/map_utilities.py
```diff
@@ -1,74 +1,296 @@
...
-        peak_theta = the_map.GetYaxis().GetBinCenter(_peakTheta.value)
-        return peak_corr, peak_phi, peak_theta
+    peak_corr = the_map.GetMaximum()
+    peak_phi = the_map.GetXaxis().GetBinCenter(_peakPhi.value)
+    peak_theta = the_map.GetYaxis().GetBinCenter(_peakTheta.value)
+    return peak_corr, peak_phi, peak_theta
 
-def calculate_avg_antenna_xyz(station_id, num_channels):
+
+def initialize_ARA_geometry(station_id):
```
marcomuzio, line 43: I think we want to make this into a new class actually, with this as one of the member functions. Probably better to avoid declaring functions inside functions.

araproc/framework/map_utilities.py
```diff
@@ -1,74 +1,296 @@
...
-        sum_z += ant_info[2]
-    
-    # Calculate averages
-    avg_x = sum_x / num_channels
-    avg_y = sum_y / num_channels
-    avg_z = sum_z / num_channels
-    
-    return avg_x, avg_y, avg_z
+    # Load required ARA libraries
+    ROOT.gSystem.Load(os.environ.get('ARA_UTIL_INSTALL_DIR') + "/lib/libAraEvent.so")
```
marcomuzio, line 56: This should already be loaded, I don't this we need this?

araproc/framework/map_utilities.py
```diff
@@ -1,74 +1,296 @@
...
-        peak_corr = the_map.GetMaximum()
-        peak_phi = the_map.GetXaxis().GetBinCenter(_peakPhi.value)
-        peak_theta = the_map.GetYaxis().GetBinCenter(_peakTheta.value)
-        return peak_corr, peak_phi, peak_theta
+    peak_corr = the_map.GetMaximum()
+    peak_phi = the_map.GetXaxis().GetBinCenter(_peakPhi.value)
+    peak_theta = the_map.GetYaxis().GetBinCenter(_peakTheta.value)
+    return peak_corr, peak_phi, peak_theta
 
-def calculate_avg_antenna_xyz(station_id, num_channels):
```
marcomuzio, line 37: I think it's okay to get ride of this if we want, but let's make sure it's not being used anywhere else in the code. In fact, maybe it's best to make this a wrapper function or something to ensure backwards compatibility.

araproc/framework/map_utilities.py
```diff
@@ -1,74 +1,296 @@
...
+                antenna_coordinates[axis].append(antenna_info.antLocation[axis])
+        av_ant_position = tuple(np.average(coords) for coords in antenna_coordinates)
+        return av_ant_position
+
+    def get_local_CP(cal_pulser_index):
+        """
+        Parameters
+        ----------
+        cal_pulser_index: int 
+            The index of calpulser (0-3)
```
marcomuzio, line 105: Maybe some more description would be helpful here. I'm guess 2 of these are the Vpols and 2 are the Hpols? Can you say which are which?

araproc/framework/map_utilities.py
```diff
@@ -1,74 +1,296 @@
...
+        global_coords = get_survey_to_global_coords(easting, northing) * FEET_TO_METER
+        global_coords = [float(global_coords[0]), float(global_coords[1]), 0]
+        st_centric_coords = get_global_to_station_centric(global_coords)
+        return st_centric_coords[0], st_centric_coords[1], pulser_depths[pulser_name]
+
+    def get_cartesian_to_spherical(coord1, coord2): 
+        """
+        Parameters
+        ----------
+        coord1: array
```
marcomuzio, line 221: What are coordinate 1 and 2 here? Seems like this is more than a conversion from cartesian to spherical? It seems like it also does a coordinate transformation?

araproc/framework/map_utilities.py
```diff
@@ -1,74 +1,296 @@
...
+        r (radius), 90-theta (elevation) and phi (azimuth)              
+        """  
+        delta = np.array(coord2) - np.array(coord1)
+        x, y, z = delta
+        r = np.sqrt(x**2 + y**2 + z**2)
+        theta = np.degrees(np.arccos(z / r))
+        phi = np.degrees(np.arctan2(y, x))
+        return r, 90 - theta, phi
+
+    return type(
```
marcomuzio, line 236: These will all be member functions of the class.

araproc/framework/map_utilities.py
```diff
@@ -1,74 +1,296 @@
...
+            "get_survey_to_global_coords": partial(get_survey_to_global_coords),
+            "get_global_to_station_centric": partial(get_global_to_station_centric),
+            "get_southpole_landmarks": partial(get_southpole_landmarks),
+            "get_distant_pulsers": partial(get_distant_pulsers),
+            "get_cartesian_to_spherical": partial(get_cartesian_to_spherical),
+        },
+    )()
+
+
+def get_known_landmarks(station_id,list_of_landmarks = None,cal_pulser_index=None,spice_depth = None):
```
marcomuzio, line 251: I think this could be a member function of the class too.

araproc/framework/constants.py
```diff
@@ -15,3 +15,5 @@
 num_samples_per_dda = 32768
 num_polarization = 2
 num_trigger_type = 3
+num_RF_channels = 16
```
marcomuzio, line 18: Given how closely tied this is to `rf_channels_ids`, probably its better to define as `num_RF_channels = len(rf_channels_ids)`

araproc/framework/map_utilities.py
pgiri4: got it!

araproc/framework/constants.py
pgiri4: Good idea!

araproc/framework/map_utilities.py
```diff
@@ -1,74 +1,296 @@
...
+    """
+    Parameters
+    ----------    
+    station_id (int): ID of the station.
+    cal_pulser_index : int (0-3)   
+      which calpulser you want to see in your skymap
+    list_of_landmarks: list
```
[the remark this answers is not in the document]
pgiri4: making this a list as input or by default one CP

araproc/framework/map_utilities.py
pgiri4: Lazy me! I am replacing coords 1 and 2 as reference_point and target_point. True! It's not simply a conversion. I can think of this function name : relative_cartesian_to_spherical? Do you have any suggestion?

araproc/framework/map_utilities.py
marcomuzio: Right outside the class. Probably it would just define an instance of the class the only exists inside the function or something

araproc/framework/map_utilities.py
marcomuzio: hmm maybe `get_relative_position` is enough? or `get_relative_spherical_position`

marcomuzio:
Thanks for adding the examples! They look really nice. Maybe we can add a label for the blue line too? Something like "SP direction" maybe?

pgiri4:
image

araproc/framework/map_utilities.py
```diff
@@ -1,74 +1,309 @@
...
+        av_ant_position = tuple(np.average(coords) for coords in antenna_coordinates)
+        return av_ant_position
+
+
+    def get_local_CP(self, list_of_cal_pulser_indices):
+        """
+        Parameters
+        ----------
+        list_of_cal_pulser_indices: list
+            The indices of calpulser [0,1,2,3] Note: 0 and 2 for HPols and 1 and 3 are VPols
```
marcomuzio, line 99: And [0,1] are on the same string I guess?

araproc/framework/data_visualization.py
```diff
@@ -117,7 +117,7 @@ def plot_waveform_bundle(
     plt.close(fig)
     del fig, axd
 
-def plot_skymap(the_map = None,
+def plot_skymap(st,list_of_landmarks = None,cal_pulse_index = None,spice_depth = None,the_map = None,
```
marcomuzio, line 120: Can you add documentation for this function?

marcomuzio: Like the usual description, parameters, returns stuff

araproc/framework/map_utilities.py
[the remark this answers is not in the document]
marcomuzio: Sorry I'm not sure I understand? I just mean, it probably would be convenient to be able to just say you want all landmarks for a station without needing to list them all one by one.

araproc/framework/map_utilities.py
pgiri4: Yes, 0 and 1 are Hpol and Vpol in one and 2 and 3 are Hpol and Vpol in another string

araproc/framework/data_visualization.py
pgiri4: sure!

araproc/analysis/standard_reco.py
```diff
@@ -579,7 +579,7 @@ def get_surface_corr_max(self, corr_map, z_thresh=-10):
         radius = float(radius)
 
         # Calculate average antenna z coordinates
-        _, _, avg_z = mu.calculate_avg_antenna_xyz(self.station_id, self.num_channels)
+        _, _, avg_z = mu.initialize_ARA_geometry(self.station_id).get_station_center(excluded_channels = [])
```
marcomuzio, line 582: Does this still work? Probably needs to be updated to use the class now?

araproc/framework/data_visualization.py
```diff
@@ -117,16 +117,35 @@ def plot_waveform_bundle(
...
                 output_file_path = None
                 ):
+    """
+    This function returns skymap for given map type
+    Parameters
+    ----------
+    list_of_cal_pulser_indices : list ## example [0,1,2,3]
+       which calpulsers you want to see in your skymap
+    list_of_landmarks: list ## example ['ICL',IC22S']
```
marcomuzio, line 129: Sorry to be a pain, but I know this will be easier for the user: can you list all valid landmark options here?

pgiri4: No worries!

pgiri4:
image
How the angle is being calculated? 
image

araproc/framework/map_utilities.py
```diff
@@ -1,74 +1,334 @@
...
+        """
+        Returns
+        -------
+        critical angle in terms of elevation angle
+
+        """
+        station_depth = abs(self.get_station_center()[2])
+        n_air = 1.0
+        n_ice = 1.78 - 0.454*np.exp(-0.0202*station_depth)
+        print('station depth in meter', station_depth, n_ice)
```
marcomuzio, line 271: Let's just drop this print statement

araproc/framework/map_utilities.py
```diff
@@ -1,74 +1,334 @@
...
+        critical_angle = np.arcsin(n_air/n_ice)
+        critical_angle *= (180/np.pi)
+        print(critical_angle)
```
marcomuzio, line 274: and this

marcomuzio:
Thanks @pgiri4 this looks perfect! We should open an issue to add the index of refraction look up at some point but I think this is perfect for the time being. Just one question: what's with CP1 in the example above? It's directly above station center?

pgiri4:
> Thanks @pgiri4 this looks perfect! We should open an issue to add the index of refraction look up at some point but I think this is perfect for the time being. Just one question: what's with CP1 in the example above? It's directly above station center?

Thanks Marco. Because A4 has no second CP and they are (0,0,0) in the CalPulser table (if you see the sql table for A4).

araproc/framework/map_utilities.py
pgiri4: Done!

marcomuzio:
Ah got it. Okay as long as it's not a bug! Merge away! Thanks Pawan!!

pgiri4:
Thanks Marco!

PR #76, opened 2025-01-06
Add ability to access raw atri station events

clark2668:
This PR enables AraProc to give the user raw atri station events, to not have to load calibration data, which makes AraProc faster if all the user wants to do is e.g. count number of cal pulsers or software triggers, etc.

### DIFF
--- a/araproc/framework/dataset.py
+++ b/araproc/framework/dataset.py
@@ -123,7 +123,7 @@ class DataWrapper:
     do_not_calibrate: bool
         a boolean that controls if the calibration (pedestals, voltage ad timing cal files) will be loaded and applied
         If True, then you can only access "raw" events, meaning no waveforms.
-        If False, then AraProc will load the pedestals, and timing/volt cal, and you will have acces to waveforms.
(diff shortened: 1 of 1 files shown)


araproc/framework/dataset.py
```diff
@@ -120,6 +120,11 @@ class DataWrapper:
         This will be inferred from the data itself
     station_id: 
         station id from 1->5, 100 (only A1-A5 supported for now)
+    do_not_calibrate: bool
+        a boolean that controls if the calibration (pedestals, voltage ad timing cal files) will be loaded and applied
+        If True, then you can only access "raw" events, meaning no waveforms.
+        If False, then AraProc will load the pedestals, and timing/volt cal, and you will have acces to waveforms.
```
marcomuzio, line 126: Maybe we should just describe calibration as `kLatestCalib`? I think that's the only option we use right? Just to be clear what is meant by timing/volt cal

marcomuzio:
Thanks @clark2668, made one small comment but I think this is ready to go.

araproc/framework/dataset.py
clark2668: Yes, that is the only one AraProc supports. I made that change!

clark2668:
Made your change! Merging and closing now!

PR #77, opened 2025-01-07
Different AraSim and AraRoot path variables

marcomuzio:
Updated `__init__.py` to have different environmental variables to give the paths to the compiled AraSim and AraRoot libraries. New variables `ARA_ROOT_LIB_DIR` and `ARA_SIM_LIB_DIR` are meant to be descriptive and direct enough that unexpected directory structures can be accommodated without directly modifying `__init__.py`.

Users will need to add these variables to their environment for AraProc to work properly.

### DIFF
--- a/araproc/__init__.py
+++ b/araproc/__init__.py
@@ -4,7 +4,7 @@ import os
 
 # libAraEvent
 try:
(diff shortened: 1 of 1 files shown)


clark2668:
Does this mean we should push a change to cvmfs as well?

marcomuzio:
CVMFS was updated in [PR#15](https://github.com/ara-software/cvmfs/pull/15)

PR #78, opened 2025-01-10
A1 cw filters

Mohammad-Neutrino:
A1 sees three most prominent CWs at 123, 224 and 403 MHz. I added three filters in narrow band to remove those (3 filters). Calibration seasons (and also sporadically in Polar winter) have frequencies in 147, 152, 275, 280, 475, 807 MHz. I clustered the other in 3 filters. Last one filter was added for entire bandwidth (50-950 MHz) to catch and remove any missed CWs. Shortening the list further is hard as the min_pow_ratio is hard to estimates for a larger bandwidth. In general, A1 being shallower, sees more CWs!

### DIFF
--- a/araproc/framework/config_files/analysis_configs.yaml
+++ b/araproc/framework/config_files/analysis_configs.yaml
@@ -2,27 +2,25 @@ station100:
   config1:
     excluded_channels : []
     filters:
-      filt1 : { min_freq : 0.122, max_freq : 0.126, min_power_ratio : 0.05 }
-      filt2 : { min_freq : 0.170, max_freq : 0.180, min_power_ratio : 0.05 }
+      filt1 : { min_freq : 0.122, max_freq : 0.126, min_power_ratio : 0.02 }
+      filt2 : { min_freq : 0.140, max_freq : 0.160, min_power_ratio : 0.01 }
       filt3 : { min_freq : 0.223, max_freq : 0.227, min_power_ratio : 0.02 }
-      filt4 : { min_freq : 0.245, max_freq : 0.255, min_power_ratio : 0.02 }
-      filt5 : { min_freq : 0.270, max_freq : 0.280, min_power_ratio : 0.02 }
-      filt6 : { min_freq : 0.396, max_freq : 0.410, min_power_ratio : 0.02 }
-      filt7 : { min_freq : 0.470, max_freq : 0.480, min_power_ratio : 0.02 }
-      filt8 : { min_freq : 0.570, max_freq : 0.580, min_power_ratio : 0.02 }
-      filt9 : { min_freq : 0.50, max_freq : 0.950, min_power_ratio : 0.10 }
+      filt4 : { min_freq : 0.245, max_freq : 0.280, min_power_ratio : 0.02 }
+      filt5 : { min_freq : 0.396, max_freq : 0.410, min_power_ratio : 0.02 }
+      filt6 : { min_freq : 0.470, max_freq : 0.580, min_power_ratio : 0.02 }
+      filt7 : { min_freq : 0.50, max_freq : 0.950, min_power_ratio : 0.10 }
     readout_limits:
       rf_readout_limit: 20
       soft_readout_limit: 8
   config2:
     excluded_channels : []
     filters:
-      filt1 : { min_freq : 0.122, max_freq : 0.126, min_power_ratio : 0.05 }
-      filt2 : { min_freq : 0.170, max_freq : 0.180, min_power_ratio : 0.05 }
+      filt1 : { min_freq : 0.122, max_freq : 0.126, min_power_ratio : 0.02 }
+      filt2 : { min_freq : 0.140, max_freq : 0.160, min_power_ratio : 0.01 }
       filt3 : { min_freq : 0.223, max_freq : 0.227, min_power_ratio : 0.02 }
-      filt4 : { min_freq : 0.245, max_freq : 0.255, min_power_ratio : 0.02 }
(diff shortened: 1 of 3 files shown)


Mohammad-Neutrino:
It seems this PR has also picked up one of my other commits on making skymap better. I made a few changes in the data_visualization's plot_skymap function for following reasons:
1. the landmarks' texts were making the map too busy especially longer texts like 'critical angle' and 'SP direction'. I changed them to '\theta_c' and '\phi_{SP}' respectively. 
2. white color texts especially for ICL was not visible for reflected maps in A1 as the upper half map was white. So I changed the color to magenta.
3. viridis base color in lower correlation is dark blue. keeping the SP direction line 'blue dashed' was making the line blend in. I changed it to cyan color for better popped up visualization
4. axes labels were specifically mentioned as elevation and azimuth to avoid confusion. symbol for degree is also added. title was changed accordingly.

Before these changes a map would look like this

[pulser_v_station_100_run_3173_event_40886_skymap_before.pdf](https://github.com/user-attachments/files/18371569/pulser_v_station_100_run_3173_event_40886_skymap_before.pdf)

With above changes, it looks like this

[pulser_v_station_100_run_3173_event_40886_skymap.pdf](https://github.com/user-attachments/files/18371573/pulser_v_station_100_run_3173_event_40886_skymap.pdf)

araproc/framework/config_files/analysis_configs.yaml
```diff
@@ -2,43 +2,39 @@ station100:
...
+      filt1 : { min_freq : 0.122, max_freq : 0.126, min_power_ratio : 0.02 }
+      filt2 : { min_freq : 0.140, max_freq : 0.160, min_power_ratio : 0.01 }
       filt3 : { min_freq : 0.223, max_freq : 0.227, min_power_ratio : 0.02 }
-      filt4 : { min_freq : 0.245, max_freq : 0.255, min_power_ratio : 0.02 }
-      filt5 : { min_freq : 0.270, max_freq : 0.280, min_power_ratio : 0.02 }
-      filt6 : { min_freq : 0.396, max_freq : 0.410, min_power_ratio : 0.02 }
-      filt7 : { min_freq : 0.470, max_freq : 0.480, min_power_ratio : 0.02 }
-      filt8 : { min_freq : 0.570, max_freq : 0.580, min_power_ratio : 0.02 }
-      filt9 : { min_freq : 0.50, max_freq : 0.950, min_power_ratio : 0.10 }
+      filt4 : { min_freq : 0.245, max_freq : 0.280, min_power_ratio : 0.02 }
```
marcomuzio, line 8: Is there a reason we can't combine filt4, 5, and 6? They are close together, without any other filters between them, and all have the same min power ratio. So why not a filter from 0.245-0.580 GHz with min power ratio 0.02?

marcomuzio:
Thanks for this update @Mohammad-Neutrino! I had one question about further combining filters. Maybe it would be helpful if you could benchmark the average time it takes to process an event with all these filters vs none.

Updates to the skymaps look nice, thanks!

araproc/framework/config_files/analysis_configs.yaml
Mohammad-Neutrino: Yes, I thought about this and tried implementing this when I was putting these filters together. Radiosonde @403 MHz has a large power. If an event has both radiosonde and CW in filter 6 bandwidth, the later is not removed! The same with filter 4 and radiosonde. If there is a large power (more than radiosonde) in ~255 MHz (A1 has many events with this CW), if we choose a filter from 250-500 MHz, only the first one is removed and we still see 403 peaks in several channels.

Mohammad-Neutrino:
Thanks @marcomuzio, yes, let me check the time difference for 'filtered' (with all 7 filters) vs 'dedispersed' events.

araproc/framework/config_files/analysis_configs.yaml
marcomuzio: hmm okay but then maybe you can have a dedicated filter for the radiosonde and then one from 250-500 MHz for the smaller lines?

marcomuzio: Also can you order all your filters from highest minimum power ratio to lowest? I think they may perform better in that case.

Mohammad-Neutrino: I think, I forgot why I kept a separate filter after radiosonde frequency. It might be because of the same reason. If there is a 255 MHz CW with larger power than one at say 470 MHz, then we will have the same problem. But, let me consolidate filters 4, 5, 6 as:
filter 4: 250-500 MHz
filter 5: 396-410 MHz

for now and if we see any CW not being filtered in filter 4 BW, we can separate those. And yes, I can arrange the filters in terms of highest to lowest minimum power ratio. Thanks!

Mohammad-Neutrino:
@marcomuzio , I checked the filtering duration for an event. This is not from L2.py but from plot_event.py. 
Took 5.336700916290283 seconds for _dedispersed_ trace.
Took 5.9189722537994385 seconds for _filtered_ trace.

marcomuzio:
Was there any improvement ordering the filters in terms of being able to consolidate them? Also the event you used for that benchmark: was it CW contaminated?

Mohammad-Neutrino:
@marcomuzio, yes, the event I checked had CW contamination. It was visibly not obvious that ordering or consolidating made any improvement. But now that I added a CW band 140-160 MHz, it filters a CW @152 MHz which seems to show up in pole seasons.

marcomuzio:
Okay good to know. Thanks for checking that! Please go ahead and merge.

Mohammad-Neutrino:
Merging now. Thanks @marcomuzio !

PR #79, opened 2025-01-10
Add fractional power axis

marcomuzio:
Added a secondary axis for AraProc generated spectrum plots which shows the fractional power. This should help make it easier to efficiently set the minimum power ratio threshold for CW filters.

Screenshot 2025-01-10 at 6 13 19 PM

### DIFF
--- a/araproc/framework/data_visualization.py
+++ b/araproc/framework/data_visualization.py
@@ -127,6 +127,7 @@ def plot_waveform_bundle(
           ax2.set_ylim(fmin, fmax)
           ax2.set_yscale('log')
           ax2.set_yticks(10**np.arange(np.ceil(np.log10(fmin)), 0+0.1, 1))
+          ax2.axhline(y=1.0, c='lightgray', linestyle='--')
           if ch in ["ch3", "ch7", "ch11", "ch15"]:
             ax2.set_ylabel("Fractional Power")
       else:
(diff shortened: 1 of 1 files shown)


clark2668:
Looks good to me. Does it make sense to add a flag for drawing the 100% horizontal line?

araproc/framework/data_visualization.py
```diff
@@ -86,6 +87,9 @@ def plot_waveform_bundle(
             # np.abs(spectrum)**2 makes spectrum in mV^2. For power, you do P = V^2/R , here R = Z_0 = 50 Ohm.
             # mV**2/50 =  mW * 1e-3. Power is always reperesented as dBm in dB scale which is 10*log10(P in mW)
             yvals = 10*np.log10(np.abs(spectrum)**2 / 50 / 1e3) # from mV to dBm
+            
+            # save total of square of absolute spectrum for reference
+            spec2Tot[f"ch{wave_key}"] = (np.abs(spectrum)**2).sum()
```
clark2668, line 92: Do you only want to calculate this if they ask for frequency domain?

marcomuzio: I think so? Is there a reason to calculate it any case?

clark2668: No, I think we should only calculate this if they ask for a freq domain representation.

marcomuzio:
I added the 100% horizontal line to always be plotted, since it's not super distracting and to avoid an extra option that doesn't make much sense in the time domain case. If you'd rather I add a more general "draw a horizontal line on the plot" functionality, let me know.

clark2668:
Nope, I think this is good. Merge away.

araproc/framework/data_visualization.py
marcomuzio: Perfect, then I think it's all good!

PR #80, opened 2025-01-11
Pair-wise plane wave reconstruction

marcomuzio:
This PR adds functionality to calculate the elevation angle of a signal by comparing the time differences between signals on a single string assuming a plane wave approximation. A combined elevation angle from all pairs is calculated by minimizing the error between each pair's reconstruction.

Also updated the skymap plotting to be able to include this and cleaned up some code.


clark2668:
Hey marco -- I think this all looks alright. Thanks!

marcomuzio:
Thanks!

PR #81, opened 2025-01-17
Fix get_corr_snr to handle empty corr. thresh cases

drparamita:
Changed the  get_corr_snr to check if index above threshold is empty. If yes, just return 0

### DIFF
--- a/araproc/analysis/standard_reco.py
+++ b/araproc/analysis/standard_reco.py
@@ -886,6 +886,10 @@ class StandardReco:
         # trim correlation function 
         corr_thresh = 1e-3
         idxAboveThresh = np.squeeze(np.where(np.abs(corr_func) >= corr_thresh)) # all indices above threshold
+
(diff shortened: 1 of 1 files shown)


araproc/analysis/standard_reco.py
```diff
@@ -886,6 +886,10 @@ def get_corr_snr(self, ch1, ch2, wave_packet):
         # trim correlation function 
         corr_thresh = 1e-3
         idxAboveThresh = np.squeeze(np.where(np.abs(corr_func) >= corr_thresh)) # all indices above threshold
+
+        # checking if idxAboveThresh is empty, if yes, just return 0
```
marcomuzio, line 890: Maybe change to: `# check if there are no above threshold indices, if so return 0`

marcomuzio:
Thanks @drparamita, made one small comment. Please update that and then merge. Thanks!

PR #82, opened 2025-01-17
Fixed some ROOT nonsense.

marcomuzio:
Redirecting output back to stdout and stderr was crashing for some reason. This should fix it. Not sure why it didn't always seem to break. 

Reference: [https://root-forum.cern.ch/t/how-to-invoke-gsystem-redirectoutput-0/14101](https://root-forum.cern.ch/t/how-to-invoke-gsystem-redirectoutput-0/14101)


clark2668:
Looks good to me

PR #83, opened 2025-01-17
Housekeeping

clark2668:
This makes AraProc able to read hk file and daq config files.

### DIFF
--- a/araproc/framework/dataset.py
+++ b/araproc/framework/dataset.py
@@ -3,46 +3,17 @@ import copy
 import ctypes
 import logging
 import numpy as np
-import os
 import ROOT
 import yaml
 
 from araproc.framework import waveform_utilities as wu
 from araproc.framework import constants as const
+from araproc.framework import file_utilities as futil
 from araproc.analysis import dedisperse as dd
 from araproc.analysis import cw_filter as cwf
 
 import importlib.resources as pkg_resources
 from . import config_files
-
-def file_is_safe(file_path):
-    
-    is_safe = False
-
-    if not isinstance(file_path, str):
-        logging.error(f"Path to file must be a string. Is {type(file_path)}")
-    if not os.path.exists(file_path):
-        logging.error(f"File ({file_path}) not found")
-    if not os.path.isfile(file_path):
-        logging.error(f"{file_path} is not a file")
-    else:
-        is_safe = True
-    
-    return is_safe
-
-def get_cvmfs_ped_file_name(station_id, run_number):
-
-    if station_id not in const.valid_station_ids:
-        raise Exception(f"Station id {station_id} is not supported")
-
-    if not np.isfinite(run_number) or run_number < 0:
-        raise Exception(f"Run number {run_number} is not supported")
-    
-    cvmfs_top_dir = "/cvmfs/icecube.osgstorage.org/icecube/PUBLIC/groups/arasoft/pedestals_v2"
-    start = start = run_number - (run_number % 1000)
-    stop = start + 999
-    file=f"station_{station_id}/{start:07d}-{stop:07d}/station_{station_id}_run_{run_number:07d}.gz"
-    return os.path.join(cvmfs_top_dir, file)
 
 
 def get_filters(station_id, analysis_config):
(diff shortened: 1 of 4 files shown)


araproc/framework/dataset.py
```diff
@@ -3,47 +3,18 @@
 import ctypes
 import logging
 import numpy as np
-import os
 import ROOT
 import yaml
 
 from araproc.framework import waveform_utilities as wu
 from araproc.framework import constants as const
+from araproc.framework import file_utilities as futil
```
marcomuzio, line 11: lol "futile"... pretty much haha

marcomuzio:
All looks good to me. Only suggestion might be an "is_calpulser_on" function in the `ConfigFileWrapper` class to easily parse that data? But maybe this isn't the place to do it. Anyway as you like. Merge when you are ready.

clark2668:
That's a good idea. How does that look?

araproc/framework/housekeeping.py
```diff
@@ -230,6 +230,7 @@ def __init__(self,
             raise Exception(f"{path_to_config_file} has a problem!")
         
         self.cal_pulser_info = self.parse_cal_pulser_info()
+        self.is_cal_pulser_on = bool(self.cal_pulser_info["opIceA"] and self.cal_pulser_info["opIceA"])
```
marcomuzio, line 233: This seems like a typo? is it supposed to be A and B?

marcomuzio: Those values are booleans I guess?

clark2668: I think you actually want A or B. (They are 1's and 0's)

araproc/framework/file_utilities.py
```diff
@@ -29,4 +30,35 @@ def get_cvmfs_ped_file_name(station_id, run_number):
...
+    # always look for the 100pct file
+    file_path = os.path.join("/data/exp/ARA/", str(year), 
+                                f"L1/100pct/ARA0{dataset.station_id}")
+    
+    sensor_hk_file = None
+    event_hk_file = None
+
+    for root, dirs, files in os.walk(file_path):
+        for file in files:
+            if str(dataset.run_number) in file:
```
marcomuzio, line 58: I think this won't work in general? Maybe it does if its expanded to be 6 digits or whatever?

clark2668: Why not? Can you share the case you're thinking about where this doesn't work?

araproc/framework/file_utilities.py
```diff
@@ -29,4 +30,35 @@ def get_cvmfs_ped_file_name(station_id, run_number):
...
+                if "sensorHk" in file:
+                    sensor_hk_file = os.path.join(root, file)
```
marcomuzio, line 60: I'm not super familiar with os.walk. `root` includes the full path except for the file? Or you need to join `dir`? as well?

clark2668: That's right, root includes the full path except for the file. So root might be: `/data/exp/ARA/2019/L1/100pct/ARA05/0701/run005626` and file might be `event005626.root`. So you end up with a totally valid path `/data/exp/ARA/2019/L1/100pct/ARA05/0701/run005626/event005626.root`

araproc/framework/file_utilities.py
marcomuzio: maybe im wrong. but if you have `run_number = 6` won't this if be true for any run with a 6 in it?

araproc/framework/file_utilities.py
marcomuzio: okay perfect. thanks!

araproc/framework/file_utilities.py
clark2668: oh shoot, I see your point. Let me think...

araproc/framework/file_utilities.py
```diff
@@ -29,4 +30,35 @@ def get_cvmfs_ped_file_name(station_id, run_number):
...
+
+    """
+    A utility function to guess the location of the housekeeping file
+    associated with a run.
+    You need to be at the WIPAC datawarehosue for this to work.
+    """
+
+    # get the year from the date of the first event in the file
+    t_start = dataset.get_raw_event(0).unixTime
+    timestamp = datetime.datetime.fromtimestamp(t_start)
```
marcomuzio, line 45: I think we don't need to do this.

araproc/framework/file_utilities.py
marcomuzio: i think if you do `str('%06d' % run_number)` is should be okay?

marcomuzio:
Okay I think this looks good!

clark2668:
Thank you! Merging....

PR #85, opened 2025-01-22
Crop waveforms functions

marcomuzio:
Adds `cropped` waveform type, whose use case is mostly to separate D & R signals. The cropping parameters are passed by users through the `crop_times` argument, which just zeros out the trace outside the passed window. This waveform type is the last to be applied in the layers of waveform processing in dataset.

### DIFF
--- a/araproc/framework/dataset.py
+++ b/araproc/framework/dataset.py
@@ -817,7 +817,8 @@ class AnalysisDataset:
 
     def get_wavepacket(self,
                       useful_event,
-                      which_traces = "filtered"
+                      which_traces = "filtered",
+                      crop_times = None,
                       ):
                      
         """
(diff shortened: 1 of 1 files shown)


araproc/framework/dataset.py
```diff
@@ -852,6 +853,14 @@ def get_wavepacket(self,
                     They are then bandpass filtered to remove out of band noise.
                     Most people should use the "filtered" traces,
                     and so they are the default argument.
+                "cropped" :
+                    These are filtered waveforms that are cropped to a time range
+                    passed by the user through the crop_times argument. Outside this
+                    range the waveform is set to 0. 
+        crop_times : dict of float pairs
```
Mohammad-Neutrino, line 860: can we add something like,
keys: (int) channel number
values: (tuple) time range

Mohammad-Neutrino:
Hi @marcomuzio this looks great! I tested it and it works. Here is an example of a channel

A2_run5505_event39072_channel5

I left one minor comment. I think after that it's good to be merged.

araproc/framework/dataset.py
marcomuzio: Yes, done!

marcomuzio:
Thanks @Mohammad-Neutrino!

PR #86, opened 2025-01-27
Quick fix on event quality issue ( if event has not known trigger type, tag as bad event)

pgiri4:
image
It will now skip events with unknown trigger type.

### DIFF
--- a/araproc/analysis/daq_quality_cut.py
+++ b/araproc/analysis/daq_quality_cut.py
@@ -55,6 +55,7 @@ def process_event_info(useful_event, station_id):
         trig_type = 0
     else:
        logger.error("__________________ Unknown trigger type for the event _______________________")  # error message
+       trig_type = None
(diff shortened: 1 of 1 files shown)


marcomuzio:
Thanks for this fix @pgiri4 I think it looks good. Just to be sure I understand what actually caused the crash when this happened? It wasn't the error statement in the logger I guess?

pgiri4:
The issue was undefined trig_type and this line https://github.com/ara-software/AraProc/blob/7b8e278fe2fb4d0d750f489c2b7fd78a06ae8380/araproc/analysis/daq_quality_cut.py#L61 was terminating the whole L2 processing of the run. Now, I assign trig_type = **None** to those events and tag them as bad events so the L2 processing of the run will not be terminated.

marcomuzio:
Ah right, duh. That makes sense. Okay then I think we are good to merge! Thanks!

PR #87, opened 2025-01-28
make futil more general, and, expose the dataset wrapper to the publi…

clark2668:
…c so I can use it elsewhere


marcomuzio:
Looks good to me. Merge it.

PR #89, opened 2025-01-29
making FFT plots less busy

Mohammad-Neutrino:
Made some changes for sharing second y-axes share ticks in FFT plots. 
previously

currently

### DIFF
--- a/araproc/framework/data_visualization.py
+++ b/araproc/framework/data_visualization.py
@@ -106,7 +106,8 @@ def plot_waveform_bundle(
         ax.set_xlabel(xlabel_options[time_or_freq])
     for ax in [axd["ch0"], axd["ch4"], axd["ch8"], axd["ch12"]]:
         ax.set_ylabel(ylabel_options[time_or_freq])
-    
+   
+    frac_pow_axes = [] 
     # make limits look nice
     for ch in axd:
       if time_or_freq == "freq":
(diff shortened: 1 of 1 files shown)


clark2668:
Sorry, what does this do? Just makes all secondary axes have the same range?

marcomuzio:
Sorry can you embed those as png's so it's easier to compare?

Mohammad-Neutrino:
@clark2668 yes, we now have two y-axes. It just makes the second y-axes share ticks and ranges.

marcomuzio:
Anyway I think we can't do this. I agree it's busy but the second y-axis scale is different for each channel. They are all in different units, so I don't think this works.

marcomuzio:
I don't think that that's true. The formula to convert the spectral amplitude to fractional power is different for each channel. If you force all the secondary y-axes to be the same, then they won't correspond the primary y-axes or the curve properly anymore.

Mohammad-Neutrino:
most_recently
 
@marcomuzio how about now?

marcomuzio:
I think this is confusing. It still implies the y-axes on the far right are shared for the entire row, which they are not. The impression is that chan 0 as a minimum at 10^-8, but that's not true.

Mohammad-Neutrino:
@marcomuzio can we associate your PR (https://github.com/ara-software/AraProc/pull/91) here?

Mohammad-Neutrino:
I still think PR 91 is getting wrong secondary y-axis ratio wrong from line 120 `y1min, y1max = axd[ch].get_ylim()`?

marcomuzio:
I'm happy to make these plots less busy in some way if you can find a way that makes sure they are still both accurate and useful. But I am pretty confident that the way the axis limits are being set is incorrect. I am pretty positive that setting the secondary y-axis limits based on `y1min, y1max = axd[ch].get_ylim()` is the correct this to do.

Mohammad-Neutrino:
I think you are right!! I was not thinking straight. I get it now. I think you wanted to keep both the y axes dependent and the graph to be represented by both.

PR #90, opened 2025-01-30
Update SineSubtract peak finding option

marcomuzio:
Old:
> Changed how SineSubtract determines which frequency has the peak power to simply use the global maximum (`GLOBALMAX` mode).

New: 
Changed how SineSubtract determines which frequency has the peak power to require that the peak be far enough over a baseline computed by a Savitzky Golay filter (`SAVGOLSUB` mode).

The rest of the discussion compares `NEIGHBORFACTOR` to `GLOBALMAX`. 

The default setting is `NEIGHBORFACTOR` which requires the candidate peak frequency to have an amplitude `1/0.15` times larger than either of its neighbors. If no candidate peak frequency satisfies this condition, the CW filtering terminates. This is a sort of safety measure to ensure it doesn't over subtract from the spectrum. However, in the case of ARA some CW features have enough width in the frequency binning that this prevents the peak from being fully subtracted. So this PR turns that feature off. However, this means that one needs to be careful that the minimum power ratios chosen to not result in over subtraction in the target band. An example is shown below.

Unfiltered waveform:
Screenshot 2025-01-30 at 4 11 29 PM

Waveform applying a 130-150 MHz filter and **minimum power ratio of 1e-9 with NEIGHBORFACTOR mode**
Screenshot 2025-01-30 at 1 50 52 PM

Waveform applying a 130-150 MHz filter and **minimum power ratio of 1e-9 with GLOBALMAX mode**
Screenshot 2025-01-30 at 4 22 49 PM

Waveform applying a 130-150 MHz filter and **minimum power ratio of 1e-3 with GLOBALMAX mode**
Screenshot 2025-01-30 at 4 28 41 PM

### DIFF
--- a/araproc/framework/dataset.py
+++ b/araproc/framework/dataset.py
@@ -60,7 +60,7 @@ def get_filters(station_id, analysis_config):
             the_filter.setVerbose(False)
             the_filter.setFreqLimits(config_settings["min_freq"], 
                                     config_settings["max_freq"])
-            the_filter.setPeakFindingOption(0) 
+            the_filter.setPeakFindingOption(0) # change to use global max as peak 
             ROOT.SetOwnership(the_filter, True) # give python full ownership
             cw_filters[filter_name] = the_filter
 
--- a/araproc/framework/dataset.py
+++ b/araproc/framework/dataset.py
@@ -60,6 +60,7 @@ def get_filters(station_id, analysis_config):
             the_filter.setVerbose(False)
             the_filter.setFreqLimits(config_settings["min_freq"], 
                                     config_settings["max_freq"])
+            the_filter.setPeakFindingOption(0) 
             ROOT.SetOwnership(the_filter, True) # give python full ownership
             cw_filters[filter_name] = the_filter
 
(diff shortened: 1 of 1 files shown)


clark2668:
This seems reasonable to me. Can we get some people to spot check the CW in addition to what you've done so far?

drparamita:
Hi @marcomuzio , at the higher frequencies, the spectrum goes down compared to the original one. Did you check why at higher frequency the spectra gets so choppy compared to the original one?

marcomuzio:
Is there a particular channel you have in mind? It's hard for me to spot this.

marcomuzio:
@clark2668 should we hold this for further tests or just merge?

drparamita:
Hi @marcomuzio, I had to wait for the run to get processed to know it CW filter with a narrow range solved the issue we are having with sine subtract algorithm alone. 
Sine subtract is clearly failing to remove the CW with the GLOBALMAX mode. So, let's merge this. I guess the example you share here are on the same event 4301, event # 18801. 
[station_5_run_4301_eventNo_18801_spectrum.pdf](https://github.com/user-attachments/files/18629541/station_5_run_4301_eventNo_18801_spectrum.pdf)
Attach is the result with sine subtract, so it does not work clearly. Please merge the PR. Thanks !

PR #92, opened 2025-01-31
Switched data file directories to temporary space.

marcomuzio:
Changed directories for timing tables and pedestals to a (hopefully) temporary location while stashcache is sorted out.


clark2668:
lgtm!

PR #93, opened 2025-02-04
catch cases where the cp file exists but is empty for some reason

clark2668:
catch cases where the cp file exists but is empty for some reason


Mohammad-Neutrino:
@clark2668 this looks good! so, for the cases where config files have cp info keys but no associated values, and if the cp trigger rate in data is low, are we flagging those runs bad?

clark2668:
Actually no -- if the file is missing, or if the key is missing, then I assume the pulser is "on." I'm making that decision because the cp information just wasn't stored in 2013 in the daq config file. Hopefully that's fine...

PR #94, opened 2025-02-07
updated CDF calculation

Mohammad-Neutrino:
Updated calculation of normalized CDF in terms of power rather than amplitude of sorted Hilbert envelope. This increases the separation in CDF related variables of signal and noise. The CDF thus changes from here

Screenshot 2025-02-06 at 1 36 57 PM

to here

Screenshot 2025-02-06 at 1 37 25 PM

### DIFF
--- a/araproc/analysis/impulsivity.py
+++ b/araproc/analysis/impulsivity.py
@@ -61,7 +61,7 @@ def get_cdf(trace):
     sorted_waveform = hilbert_envelope[clo_sort_idx]
 
     # Cumulative distribution function (CDF) calculation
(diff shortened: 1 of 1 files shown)


marcomuzio:
Looks good to me! I imagine this was used to generate the above curve?

Mohammad-Neutrino:
yes, that was the only change that went to the above plots.

marcomuzio:
sounds great, please merge! thanks @Mohammad-Neutrino this was a good catch

Mohammad-Neutrino:
Merging now. Thanks @marcomuzio !

PR #95, opened 2025-02-08
Average rms

marcomuzio:
Added a simple function to calculate the channel-average rms.


Mohammad-Neutrino:
@marcomuzio it looks good to me. Thanks for adding this!

PR #96, opened 2025-02-21
Added real run number for simulations

marcomuzio:
Changed the run number for simulations from `-10` to the run number in the passed root file. This will help with analyzing simulations and finding bugs.

The parsing is pretty simple but should work based on how the AraSim generates output file names: https://github.com/ara-software/AraSim/blob/a10009706e1d9f256f7578d0eb98c9db6a9137d6/AraSim.cc#L198 (generally simulations we produce will always have `argc > 2` when running AraSim so that the run number is part of the file name)

### DIFF
--- a/araproc/framework/dataset.py
+++ b/araproc/framework/dataset.py
@@ -428,6 +428,9 @@ class SimWrapper:
         if '.run' not in path_to_data_file:
             self.run_number = -10 # if there's no file we will crash anyway
         else:
+            # arasim files come in the form /path/AraOut.[setupFile].run[runNo].root so first
+            # split on '.run' and grab the last element to get [runNo].root then split on
+            # '.root' and grab the first element to be left with [runNo]
             self.run_number = int(path_to_data_file.split('.run')[-1].split('.root')[0])
         self.station_id = None
         self.num_events = None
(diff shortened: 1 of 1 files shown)


pgiri4:
Looks good to me.

araproc/framework/dataset.py
```diff
@@ -425,7 +425,10 @@ def __init__(self,
         self.event_tree = None
         self.sim_tree = None
         self.sim_settings_tree = None
-        self.run_number = -10 # this doesn't make sense for AraSim
+        if '.run' not in path_to_data_file:
+            self.run_number = -10 # if there's no file we will crash anyway
+        else:
+            self.run_number = int(path_to_data_file.split('.run')[-1].split('.root')[0])
```
clark2668, line 431: Can you add a comment describing what happens? I think this says, "split on .root", so you end up with an array that's like ["filename.run1234", ".root"]. That's why you grab entry 0. Then you split on run, which gives you ["filename", ".run", "1234"] (or something?), which is why you're grabbing the last element?

marcomuzio: Right, technically it happens in the other order but ya that's the idea. Just added a comment to this effect.

clark2668: Okay, then this looks good.

clark2668:
Okay, this look good to merge.

marcomuzio:
Thanks all!

PR #97, opened 2025-02-27
Add support for filtering IDed CW

marcomuzio:
This PR adds functionality for users to pass a pre-calculated ROOT file containing IDed CW for events in a run. 

When a file is passed, CW filters specified in `analysis_configs.yaml` are only applied to a given event & polarization if an identified CW frequency is in their filtering band, otherwise no filtering is applied (NB even if a filter does contain an identified frequency, the spectral amplitude in the band must still meet the minimum power ratio criteria for the filter to subtract anything).

If no file is passed, the default behaviour is to apply all CW filters as has been done previously.

Additional filter information is now tracked since this information does not have getter functions and the relevant variables are private in the `SineSubtract` class.

Some important limitations to note in the current implementation:
- Applying IDed CW filtering to simulation events it not currently supported
- When IDed CW filtering is enabled it is not possible to _also_ apply a "blanket" CW filter for all events

### DIFF
--- a/araproc/analysis/cw_filter.py
+++ b/araproc/analysis/cw_filter.py
@@ -117,6 +117,7 @@ def get_active_filters(cw_filters, cw_ids, chan):
         badFreqs = cw_ids[key]
 
         for freq in badFreqs:
+            freq /= 1e3 # convert to GHz
             for filter_i, filter in cw_filters.items():
                 # this filter is already activated
                 if filter_i in active_filters: 
@@ -130,7 +131,7 @@ def get_active_filters(cw_filters, cw_ids, chan):
                     active_filters[filter_i] = filter
 
     # filters are applied in order they appear in dict
-    # there's some advantage to apply the in order of descending 
+    # there's some advantage to apply them in order of descending 
     # min_power_ratio, so let's quickly enforce that
     active_filters = {k : v for k, v in sorted(active_filters.items(), key=lambda x: x[1]["min_power_ratio"])}
 
--- a/araproc/analysis/cw_filter.py
+++ b/araproc/analysis/cw_filter.py
@@ -117,7 +117,6 @@ def get_active_filters(cw_filters, cw_ids, chan):
         badFreqs = cw_ids[key]
 
         for freq in badFreqs:
-            freq /= 1e3 # convert to GHz
             for filter_i, filter in cw_filters.items():
                 # this filter is already activated
                 if filter_i in active_filters: 
--- a/araproc/analysis/cw_filter.py
+++ b/araproc/analysis/cw_filter.py
@@ -2,6 +2,7 @@ import logging
 import ROOT
 import yaml
 import copy
+import numpy as np
 
 from araproc.framework import constants as const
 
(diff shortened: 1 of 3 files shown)


araproc/analysis/cw_filter.py
```diff
@@ -63,7 +70,69 @@ def apply_filters(cw_filters, waveform_bundle):
...
+    else: 
+      pol = "h"
+
+    scan_directions = ["fwd", "bwd"]
+    for direction in scan_directions:
+
+        key = f"badFreqs_{direction}_{pol}"
+        badFreqs = cw_ids[key]
+
+        for freq in badFreqs:
```
clark2668, line 119: Is it worth doing a bit of sanity checking? For example, if the user passes a CW frequency larger than a few GHz, then it complains, because the units are wrong (e.g in MHz)? Also, checking for negative frequencies.

araproc/framework/dataset.py
```diff
@@ -276,6 +298,70 @@ def __load_pedestal(self, path_to_pedestal_file = None):
...
+            if futil.file_is_safe(path_to_cw_ids):
+                logging.info(f"Will try to load cw id file: {path_to_cw_ids}")
+                self.path_to_cw_ids = path_to_cw_ids
+            else:
+                raise Exception(f"{path_to_cw_ids} has a problem!")
+ 
+        else:
+            print("Automatic loading of cw id files is not yet implemented. Please pass a file path.")
+            print("Continuing without CW ID info...")
+            return
```
clark2668, line 317: Does just "returning" give the right behavior?

clark2668 (changes requested):
Can we also add an example of how to use this to the `examples` folder? But I think this looks good otherwise modulo a couple small suggestions.

araproc/analysis/cw_filter.py
marcomuzio: I thought about this a little, but I'm a bit worried about the overhead checking this for every single event? Maybe it's not so bad though. Maybe we there's an efficient way to check this with ROOT at read-in.

araproc/framework/dataset.py
marcomuzio: Should be? This function is essentially a `void` in c++ language, so returning here should just short-circuit the function, which leaves `self.cw_id_tfile`, `self.cw_id_tree`, and `self.cw_id_ptrs` as their default of `None`.

marcomuzio: Also to be sure, not passing the path to the cw file did print this message and default to the old behavior of applying all the filters.

araproc/analysis/cw_filter.py
clark2668: That's fair enough. We can not worry about it for now.

araproc/framework/dataset.py
clark2668: Okay, excellent. I guess I just wanted to make sure that the stuff you mentioned above stayed appropriately empty.

clark2668:
Okay, LGTM

clark2668 (approved):

araproc/analysis/cw_filter.py
marcomuzio: Okay nevermind, I forgot that the vector structure in the branches makes the usual ROOT functions a bit of a headache. I added a quick sanity check to the start of the cw function, let me know if you think the checks make sense/are sufficient.

marcomuzio:
Thanks @clark2668! Merging

PR #98, opened 2025-03-01
Jackknife RMS

marcomuzio:
This PR adds a sort-of jackknife estimate of the RMS of a trace and changes this to be the default RMS estimator in SNR calculations. The goal of this technique is both maximize the amount of the trace used to estimate the RMS and also be robust against impulsive signals in the trace. It does this by splitting the trace into segments and attempting to identify which segments are "outliers" in that they contain signal. It then calculates the RMS on all non-outlier segments in the trace. Segments are in terms of a fixed number of samples (to reduce sensitivity to the total trace length). By default 50 samples are used, corresponding to a 25 ns segment with our default 0.5 ns interpolation.

This new method has the advantage over our current estimate that it does not bias the RMS downwards (due to our current estimator being the `min` of segments' RMS) and results in a less dramatic difference RMS between RF triggers and software triggers (some difference is unavoidable, due to the significantly different trace lengths).

Below are some comparisons of the jackknife RMS compared to our current method, which splits the trace into `nsegs` equal-length segments and takes the minimum among these to be the RMS estimator. Comparisons are shown for `nsegs` values of 1 (the entire trace), 2, 4, 8 (default), and 16. All estimates are calculated for channel 0. 

Keep in mind that the current method always biases the RMS downwards.

**Example 1: A5 Run 5000 Event 1038 (CP trigger)**

```
nsegs = 1: 59.13
nsegs = 2: 43.91
nsegs = 4: 41.59
nsegs = 8: 39.73
nsegs = 16: 34.19
jackknife: 44.25
```

Jackknife matches well with `nsegs=2` which is the smallest number which would result in a signal-less segment.

Screenshot 2025-03-01 at 2 09 46 AM

**Example 2: A5 Run 5000 Event 1044 (software trigger)**

```
nsegs = 1: 40.41
nsegs = 2: 39.30
nsegs = 4: 37.03
nsegs = 8: 33.95
nsegs = 16: 23.28
jackknife: 39.18
```

Note that the jackknife estimate is close to the `nsegs=1` estimate (essentially the true RMS in this case) and much less biased than the default `nsegs=8` estimate. Note that even the RMS of the full trace is slightly smaller than RF triggers due to the small number of samples.

Screenshot 2025-03-01 at 2 10 44 AM

**Example 3: A5 Run 5000 Event 1049 (RF trigger)**

```
nsegs = 1: 43.78
nsegs = 2: 42.45
nsegs = 4: 40.68
nsegs = 8: 39.99
nsegs = 16: 35.29
jackknife: 41.81
```

No really obvious signal exists here and jackknife agrees roughly with `nsegs=2`.

Screenshot 2025-03-01 at 2 12 22 AM

**Example 4: A1 Run 16562 Event 8865 (RF trigger -- SPIce event)**

```
nsegs = 1: 36.92
nsegs = 2: 31.54
nsegs = 4: 20.50
nsegs = 8: 19.86
nsegs = 16: 18.99
jackknife: 24.30
```

Jackknife is slightly larger than `nsegs=4`, which is the smallest example which would result in a signal-less segment, avoiding both pulses.

Screenshot 2025-03-01 at 2 15 31 AM

**Example 5: A1 Run 16562 Event 8817 (RF trigger)**

```
nsegs = 1: 20.22
nsegs = 2: 18.67
nsegs = 4: 17.87
nsegs = 8: 17.63
nsegs = 16: 16.70
jackknife: 18.91
```

Again, no really obvious signal exists here and jackknife agrees roughly with `nsegs=2`.

Screenshot 2025-03-01 at 2 17 13 AM

### DIFF
--- a/araproc/analysis/snr.py
+++ b/araproc/analysis/snr.py
@@ -139,15 +139,17 @@ def get_jackknife_rms(waveform, nSamples=50):
     # then calculate the RMS on the trace _without_ that segment (ie the "all-but RMS")
     allRms = []
     traceLen = len(trace)
-    for i in range(0, traceLen, nSamples):
+    trace2 = np.square(trace)
+    trace2Sum = np.sum(trace2)
+    
+    start_idx = np.arange(0, traceLen, nSamples)
+    end_idx = np.clip(start_idx + nSamples, 0, traceLen)
+    segLens = end_idx - start_idx     
+    nSegs = len(segLens)
 
-      start = i
-      end = start + nSamples
-
-      subTrace = np.concatenate([trace[:start], trace[end+1:]])
-      thisRms = np.sqrt(np.mean(subTrace**2))
-      allRms.append(thisRms)
-   
+    segTrace2Sum = np.array([np.sum(trace2[start:end]) for start, end in zip(start_idx, end_idx)])
+    allRms = np.sqrt((trace2Sum-segTrace2Sum)/(traceLen - segLens))
+ 
     # segments that have outlier (ie nonthermal) voltages, will have
     # an all-but RMS that is significantly smaller than others
     # here we attempt to identify those outlier segments by detecting
@@ -155,16 +157,14 @@ def get_jackknife_rms(waveform, nSamples=50):
     # all-but RMS values by more the 1 standard deviation of the other all-but RMS values 
     # non-outlier segments are collected to calculate the final RMS
     allRms = np.asarray(allRms)
-    smoothedTrace = np.empty(0)
-    for i in range(len(allRms)):
-        subRms = np.concatenate([allRms[:i],allRms[i+1:]])
-        subMean = subRms.mean()
-        subStd = subRms.std()
-        if(allRms[i] > subMean-subStd):
-            traceSegment = trace[i*nSamples:(i+1)*nSamples]
-            smoothedTrace = np.concatenate([smoothedTrace, traceSegment]) 
+    subMean = (allRms.sum()-allRms)/(nSegs-1.) # mean of all other all-but RMS values
+    subStd = np.sqrt((np.sum(np.square(allRms-subMean)) - np.square(allRms-subMean))/(nSegs-1.)) # std of all other all-but RMS values
+    mask = np.zeros(traceLen).astype(bool)
+    for i in range(nSegs):
+        mask[i*nSamples:(i+1)*nSamples] = (allRms[i] >  subMean[i] - subStd[i]) # if all-but RMS isn't an outlier mark this segment to be included
     
-    rms = np.sqrt(np.mean(smoothedTrace**2))   
+    # calculate RMS from non-outlier segments
+    rms = np.sqrt(np.mean(trace2[mask]))   
  
     return rms
 
--- a/araproc/analysis/snr.py
+++ b/araproc/analysis/snr.py
@@ -162,7 +162,7 @@ def get_jackknife_rms(waveform, nSamples=50):
     subStd = np.sqrt((np.sum(np.square(allRms-subMean)) - np.square(allRms-subMean))/(nSegs-1.)) # std of all other all-but RMS values
     
     mask = (allRms >  subMean - subStd) # if all-but RMS isn't an outlier mark this segment to be included
-    mask = np.repeat(mask, nSamples)[:traceLen] # convert from the outlier mask array from being per-segment to per-sample
+    mask = np.repeat(mask, nSamples)[:traceLen] # convert the outlier mask array from being per-segment to per-sample
     
     # calculate RMS from non-outlier segments
     rms = np.sqrt(np.mean(trace2[mask]))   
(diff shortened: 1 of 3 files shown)


Mohammad-Neutrino (commented):
Hi Marco! This looks good to me. I think there might be some indentation issue? It can also be my system view? I will test it tomorrow on one data and one simulated run. Thank you!

araproc/analysis/snr.py
```diff
@@ -96,6 +96,76 @@ def get_min_segmented_rms(waveform, nsegs=8):
...
+    rms : float
+        The jackknifed RMS in same units as trace.
+    """
+    
+    if(isinstance(waveform, ROOT.TGraph)):
+      _, trace = wfu.tgraph_to_arrays(waveform)
+    elif(isinstance(waveform, np.ndarray)):
+      trace = np.copy(waveform)
+      
+      # don't be bothered by some unused dimensions
```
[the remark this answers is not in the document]
marcomuzio: These indents are actually as intended. It's a if, else if, else block

marcomuzio:
Thanks @Mohammad-Neutrino! The indents are actually okay, since they only apply to the case where the user passes a numpy array. You can see this in the other functions above too.

araproc/analysis/snr.py
```diff
@@ -96,6 +97,77 @@ def get_min_segmented_rms(waveform, nsegs=8):
...
+
+      if(trace.ndim != 1):
+        raise Exception("Trace is not 1d in snr.get_jackknife_rms. Abort")
+
+    else:
+      raise Exception("Unsupported data type in snr.get_jackknife_rms. Abort")
+
+    # first split the trace into segments of nSamples 
+    # then calculate the RMS on the trace _without_ that segment (ie the "all-but RMS")
+    allRms = []
```
Mohammad-Neutrino, line 140:
```suggestion
    # Split the trace into segments
    segments = [trace[i:i + nSamples] for i in range(0, len(trace), nSamples)]

    # Calculate the all-but RMS for each segment
    allRms = []
    for i in range(len(segments)):
        subTrace = np.concatenate(segments[:i] + segments[i + 1:])
        allRms.append(np.sqrt(np.mean(subTrace**2)))

    allRms = np.array(allRms)
```

something like this works better?

araproc/analysis/snr.py
```diff
@@ -96,6 +97,77 @@ def get_min_segmented_rms(waveform, nsegs=8):
...
+      allRms.append(thisRms)
+   
+    # segments that have outlier (ie nonthermal) voltages, will have
+    # an all-but RMS that is significantly smaller than others
+    # here we attempt to identify those outlier segments by detecting
+    # all-but RMS values which are smaller than the mean of the other
+    # all-but RMS values by more the 1 standard deviation of the other all-but RMS values 
+    # non-outlier segments are collected to calculate the final RMS
+    allRms = np.asarray(allRms)
+    smoothedTrace = np.empty(0)
```
Mohammad-Neutrino, line 158:
```suggestion
    # Identify non-outlier segments
    subMeans = np.array([np.mean(np.delete(allRms, i)) for i in range(len(allRms))])
    subStds = np.array([np.std(np.delete(allRms, i)) for i in range(len(allRms))])
    nonOutlierMask = allRms > (subMeans - subStds)

    # Combine non-outlier segments
    smoothedTrace = np.concatenate([segments[i] for i in range(len(segments)) if nonOutlierMask[i]])
```

And maybe this part can be vectorized like this?

Mohammad-Neutrino (commented):
Hi @marcomuzio , I have tested this and it works! However I found it adding about 1 sec per event in L2 processing. I have left some comments which makes the function more efficient and shorter. Let me know how it looks. Thanks!

marcomuzio:
Thanks for the suggestion @Mohammad-Neutrino! With some assist from chat-gpt for a clever way to vectorize this I got this down to only need 1 for loop. Unfortunately, as is usually the case, the code is a bit less readable now in my opinion. Doing a quick test this now seems to be roughly competitive with the old RMS calculation. For example, in the L2 script with the old method the full L2 processing takes ~0.4 seconds and now with jackknife it takes ~0.41 seconds (before this optimization it was closer to ~0.5 seconds).

Mohammad-Neutrino:
Thanks so much @marcomuzio for making the code faster! I think it's still readable with all the nice comments you added. Could you also, modify these:

[hilbert.py](https://github.com/ara-software/AraProc/blob/f7e28c03a6a0603d2ac580ab353bf70940ee97f9/araproc/analysis/hilbert.py#L30)
[rpr.py](https://github.com/ara-software/AraProc/blob/f7e28c03a6a0603d2ac580ab353bf70940ee97f9/araproc/analysis/rpr.py#L41)

with `get_jackknife_rms() ` function in this same PR? After this, I think it's good to get merged! Thank you!

marcomuzio:
Ah good catch! Forgot all about those. And I was able to finally get rid of those last two `for` loops, so now I think it's fully vectorized and a hair faster. Thanks for your help with this @Mohammad-Neutrino!

PR #99, opened 2025-03-12
Added hilbert peak over rms

marcomuzio:
This PR adds a calculation of the Hilbert envelope peak power divided by the rms in a portion of the envelope around the peak. This was developed on the PA CR search as an additional metric of the impulse-like nature of a trace that is less correlated with SNR. The current intent is to calculate this only for the CSW.

### DIFF
--- a/araproc/analysis/hilbert.py
+++ b/araproc/analysis/hilbert.py
@@ -72,7 +72,7 @@ def get_avg_hilbert_snr(wave_bundle, excluded_channels = []):
     return avg_hill_snr
 
 
-def get_peak_over_rms(waveform, fraction=0.5):
+def get_peak_over_avg_power(waveform, fraction=0.5):
     
     """
     Calculates the peak power at maximum of the Hilbert envelope divided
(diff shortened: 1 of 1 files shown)


Mohammad-Neutrino:
Hi @marcomuzio , thank you very much for adding this function. I am glad that I decided to take a look because this actually something that I thought about last semester when I took a class on fiber optics and another on signal processing. I think this resembles mostly with something they call "Peak-to-Average Power Ratio (PAPR)" in signal processing. I think the function's name is a bit ambiguous as we are not using RMS but rather MEAN of second half of power envelope? Can we use something like `get_papr()`. If you decide to use RMS, then there is another cool name of this quantity in fiber optics - "Normalized Peak-to-RMS Ratio (NPR)"; so, accordingly a name like `get_npr()` would be cool. 

Also, I have tested it and it works fine, so maybe after these possible changes, it's good to go. Thanks again!

marcomuzio:
Thanks @Mohammad-Neutrino! I agree, I changed the name to be more accurate and clear.

PR #100, opened 2025-03-13
Added exception in case frequency has no filter

marcomuzio:
Added a check to make sure every IDed frequency has at least one filter applied. If it doesn't, it crashes and warns the user that it needs one.


clark2668:
This looks good!

clark2668:
I wondered a bit if it makes sense to loop over freqs and then filters, or over filters and then freqs, but probably it doesn't matter.... The way you did it probably makes more sense.

PR #101, opened 2025-03-13
improved peak-to-peak calculation

Mohammad-Neutrino:
This PR makes changes in the `get_vpp` function for calculating peak-to-peak voltage. The changed function improves accuracy and flexibility by supporting both global and local extrema-based calculations. 

For a CP event
CP

For a SPIce event
SPIce

For an RF event (with signal)
RF_Pulse

For an RF event (with fluctuations)
RF_Noise

### DIFF
--- a/araproc/analysis/snr.py
+++ b/araproc/analysis/snr.py
@@ -45,11 +45,11 @@ def get_vpp(waveform, order = 1, use_local = True):
       return vpp
 
     # append 0 to the start and end of the trace to handle edge points in 'argrelextrema'
-    trace = np.pad(trace, (1, 1), mode = 'constant', constant_values = 0)
+    padded_trace = np.pad(trace, (1, 1), mode = 'constant', constant_values = 0)
 
     # find local extrema
-    upper_peak_idx = argrelextrema(trace, np.greater, order = order)[0]
-    lower_peak_idx = argrelextrema(trace, np.less, order = order)[0]
+    upper_peak_idx = argrelextrema(padded_trace, np.greater, order = order)[0]
+    lower_peak_idx = argrelextrema(padded_trace, np.less, order = order)[0]
 
     # combine and sort indices (using np.unique)
     peak_idx = np.unique(np.concatenate((upper_peak_idx, lower_peak_idx)))
--- a/araproc/analysis/snr.py
+++ b/araproc/analysis/snr.py
@@ -1,23 +1,29 @@
 import ROOT
 import numpy as np
+from scipy.signal import argrelextrema
 
 from araproc.framework import waveform_utilities as wfu
 
-def get_vpp(waveform):
+def get_vpp(waveform, order = 1, use_local = True):
 
-    """
-    Calculates the peak-to-peak voltage of a voltage trace.
+    """ 
+    Calculates the peak-to-peak voltage of a signal using local (Default) and global extrema.
 
     Parameters
     ----------
     waveform: TGraph or np.ndarray
         A TGraph or np.ndarray of the waveform voltage.
+    order : int, optional (Default = 1)
+        How many points on each side to compare for finding extrema.
+    use_local: bool, optional (Default = True)
+        If True, returns the peak-to-peak voltage based on local extrema. 
+        If False, returns the peak-to-peak voltage based on global extrema.
 
     Returns
     -------
     vpp : float
         Peak-to-peak voltage in same units as trace.
-    """ 
+    """
 
     if(isinstance(waveform, ROOT.TGraph)):
       _, trace = wfu.tgraph_to_arrays(waveform)
@@ -33,9 +39,21 @@ def get_vpp(waveform):
     else:
       raise Exception("Unsupported data type in snr.get_vpp. Abort")
 
-    vMax = trace.max()
-    vMin = trace.min()
-    vpp = vMax - vMin
+    # global extrema case
+    if not use_local:
+      vpp = trace.max() - trace.min()
+      return vpp
+
+    # find local extrema
+    upper_peak_idx = argrelextrema(trace, np.greater_equal, order = order)[0]
+    lower_peak_idx = argrelextrema(trace, np.less_equal, order = order)[0]
+
+    # combine and sort indices (using np.unique)
+    peak_idx = np.unique(np.concatenate((upper_peak_idx, lower_peak_idx)))
+
+    # compute the difference between consecutive extrema and peak directly from indices 
+    diff_peak = np.abs(np.diff(trace[peak_idx]))
+    vpp = diff_peak.max()
 
     return vpp
 
--- a/araproc/analysis/snr.py
+++ b/araproc/analysis/snr.py
@@ -4,7 +4,7 @@ from scipy.signal import argrelextrema
 
 from araproc.framework import waveform_utilities as wfu
 
-def get_vpp(waveform, order = 1, use_local = True, time_window = 20, dt = 0.5):
+def get_vpp(waveform, order = 1, use_local = True, time_window = 20, dt = 0.5, time = None):
 
     """ 
     Calculates the peak-to-peak voltage of a signal using local (Default) and global extrema.
@@ -22,6 +22,8 @@ def get_vpp(waveform, order = 1, use_local = True, time_window = 20, dt = 0.5):
         Time window in nanoseconds to calculate the peak-to-peak voltage.
     dt : float, optional (Default = 0.5 ns)
         Sampling interval in nanoseconds for synthetic time array.
+    time : np.ndarray, optional (Default = None)
+        Optional time array to use for waveform. If None, synthetic time will be created.
 
     Returns
     -------
@@ -43,9 +45,13 @@ def get_vpp(waveform, order = 1, use_local = True, time_window = 20, dt = 0.5):
       if(trace.ndim != 1):
         raise Exception("Trace is not 1d in snr.get_vpp. Abort")
 
-      # create synthetic time array based on dt = 0.5 ns
-      time = np.arange(len(trace)) * dt
-      
+      # use provided time or create synthetic time if not given
+      if time is None:
+        time = np.arange(len(trace)) * dt
+      else:
+        if len(trace) != len(time):
+          raise ValueError("Provided time array must match trace length.")
+
     else:
       raise Exception("Unsupported data type in snr.get_vpp. Abort")
 
--- a/araproc/analysis/snr.py
+++ b/araproc/analysis/snr.py
@@ -4,15 +4,15 @@ from scipy.signal import argrelextrema
 
 from araproc.framework import waveform_utilities as wfu
 
-def get_vpp(waveform, order = 1, use_local = True, time_window = 20, dt = 0.5, time = None):
+def get_vpp(waveform, order = 1, use_local = True, time_window = 20):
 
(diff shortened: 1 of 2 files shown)


marcomuzio:
Thanks for doing this @Mohammad-Neutrino, this is very promising! What are you plotting in the pink exactly? Is it the `diff_peak` array? Could you do the same as the "global" case and show the the positive and negative peaks it's using for the peak-to-peak?

araproc/analysis/snr.py
```diff
@@ -33,9 +39,21 @@ def get_vpp(waveform):
...
-    vMax = trace.max()
-    vMin = trace.min()
-    vpp = vMax - vMin
+    # global extrema case
+    if not use_local:
+      vpp = trace.max() - trace.min()
+      return vpp
+
+    # find local extrema
+    upper_peak_idx = argrelextrema(trace, np.greater_equal, order = order)[0]
```
marcomuzio, line 48: Looking at the documentation for this function, it looks like if a peak is ever at the edge of the waveform it will not be flagged as an extremum, which might not be ideal. Maybe for this purpose we should append a `0` to the start and end of the trace?

araproc/analysis/snr.py
```diff
@@ -33,9 +39,21 @@ def get_vpp(waveform):
...
+    lower_peak_idx = argrelextrema(trace, np.less_equal, order = order)[0]
```
marcomuzio, line 49: In both of these, instead of accessing the array from the tuple with `[0]` can you use `np.squeeze(argrelextrema(trace, np.less_equal, order = order))`? At first I thought you were only getting the first index and was confused. I think using `squeeze` will be more clear.

araproc/analysis/snr.py
marcomuzio: Also why do you use `np.greater_equal` and not `np.greater`?

araproc/analysis/snr.py
```diff
@@ -33,9 +39,21 @@ def get_vpp(waveform):
...
+
+    # combine and sort indices (using np.unique)
+    peak_idx = np.unique(np.concatenate((upper_peak_idx, lower_peak_idx)))
+
+    # compute the difference between consecutive extrema and peak directly from indices 
+    diff_peak = np.abs(np.diff(trace[peak_idx]))
```
marcomuzio, line 55: Is it possible that two consecutive extrema are both peaks? Or both troughs? I think it might be possible since were not using strict comparisons right now. I also wonder if noise between a peak and trough could cause an underestimate when only using consecutive extrema. It seems like this is essentially a soft condition on the nearness of the extra so maybe an explicit condition in time will be more robust?

marcomuzio (changes requested):

araproc/analysis/snr.py
Mohammad-Neutrino: Ohh, I haven't thought about this at all! You are right. It's possible for the saturating-like cases (rare)? I think then strict  `np.greater/np.less` will be better. Also a 2 samples extrema condition is better?

araproc/analysis/snr.py
marcomuzio: Hmm, I hadn't thought about that, it's a good point. Maybe using `>=` is okay to identify extrema if we change how the final difference is calculated.

araproc/analysis/snr.py
marcomuzio: Maybe here we should just take the max difference between extrema within some time window?

marcomuzio:
One other thing that occurred to me: in the case that something went wrong and a trace is exactly `0`, what is the behavior? Will it crash the code? Does it identify every sample as an extremum or none of them? I could it being an issue if none are returned.

araproc/analysis/snr.py
Mohammad-Neutrino: okay, I'll use np.squeeze then.

araproc/analysis/snr.py
Mohammad-Neutrino: okay, let me make it 5 samples window.

marcomuzio: Just to record it here: on a Zoom call we discussed using a 20 ns window, based on the impulse response of the antenna + a safety margin.

marcomuzio:
Ya I think that would be a reasonable way to handle it!

marcomuzio (changes requested):
Thanks @Mohammad-Neutrino looks like this is almost there! We should make sure to check that this isn't significantly slower than current version. Probably it's not, since this is already vectorized pretty well I think.

araproc/analysis/snr.py
```diff
@@ -29,13 +42,47 @@ def get_vpp(waveform):
...
+
+    # find local extrema
+    upper_peak_idx = np.squeeze(argrelextrema(padded_trace, np.greater, order = order))
+    lower_peak_idx = np.squeeze(argrelextrema(padded_trace, np.less, order = order))
+
+    # combine and sort indices (using np.unique)
+    peak_idx = np.unique(np.concatenate((upper_peak_idx, lower_peak_idx)))
+
+    # if there is no peak
+    if len(peak_idx) == 0:
```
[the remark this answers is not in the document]
Mohammad-Neutrino: I think, I thought adding this also a check if there is no peak but maybe this is redundant as I added this for the time-windowed index here?

```
# if there is not enough peaks in the time window
    if len(valid_peak_idx) < 2:
      return 0.0
```

araproc/analysis/snr.py
```diff
@@ -29,13 +42,47 @@ def get_vpp(waveform):
 
       if(trace.ndim != 1):
         raise Exception("Trace is not 1d in snr.get_vpp. Abort")
+
+      # create synthetic time array based on dt = 0.5 ns
+      time = np.arange(len(trace)) * dt
```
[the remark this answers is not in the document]
Mohammad-Neutrino: I actually tested it passing a ndarray trace only and it works fine. I can pass time = none optional argument in the function if you want but. I think for vpp at least it won't raise error doing this way.

araproc/analysis/snr.py
```diff
@@ -29,13 +42,47 @@ def get_vpp(waveform):
...
+
+    # if there is no peak
+    if len(peak_idx) == 0:
+      return 0.0
+
+    # adjust indices to account for the padding
+    peak_idx = peak_idx - 1
+
+    # filter peak indices within the specified time window
+    time_diff = time[peak_idx] - time[peak_idx[0]]
```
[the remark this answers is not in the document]
Mohammad-Neutrino: yes, maybe I should specify that in comment.

araproc/analysis/snr.py
```diff
@@ -29,13 +42,47 @@ def get_vpp(waveform):
...
+    # if there is no peak
+    if len(peak_idx) == 0:
+      return 0.0
+
+    # adjust indices to account for the padding
+    peak_idx = peak_idx - 1
+
+    # filter peak indices within the specified time window
+    time_diff = time[peak_idx] - time[peak_idx[0]] 
+    valid_peak_idx = peak_idx[time_diff <= time_window]
```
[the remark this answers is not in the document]
Mohammad-Neutrino: Ohh right!!! This one is definitely the right one since it searches over all valid extrema pairs. I will make the changes. I think np.flatten() does not exist but ndarray.flatten() does which I will switch to.

araproc/analysis/snr.py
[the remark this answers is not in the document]
Mohammad-Neutrino: with the suggested changes, I think, < 2 makes more sense since it handles both the case of zero and single extrema cleanly.

Mohammad-Neutrino:
@marcomuzio thanks for the suggestions! I have made the necessary changes.

araproc/analysis/snr.py
```diff
@@ -1,26 +1,41 @@
...
-    """ 
+    """
 
     if(isinstance(waveform, ROOT.TGraph)):
-      _, trace = wfu.tgraph_to_arrays(waveform)
+      time, trace = wfu.tgraph_to_arrays(waveform)
+      if len(trace) != len(time):
+        raise ValueError("Waveform and time must have the same length.")
+
     elif(isinstance(waveform, np.ndarray)):
```
marcomuzio, line 39: I think the alternative here to a TGraph should be a tuple or list. Let's do something like this:

```
elif(isinstance(waveform, list) or isinstance(waveform, tuple)):
  if len(waveform) != 2:
    raise ValueError("waveform tuple or list must have length 2!")
    
  time, trace = waveform
  if( (not isinstance(time, np.ndarray)) or (not isinstance(trace, np.ndarray)) ):
    raise ValueError("waveform must be tuple or list of np.ndarrays!")
```

then when this is called we will pass these arrays as `get_vpp(waveform=(time, trace))`

araproc/analysis/snr.py
```diff
@@ -29,13 +44,57 @@ def get_vpp(waveform):
 
       if(trace.ndim != 1):
         raise Exception("Trace is not 1d in snr.get_vpp. Abort")
-      
+
+      # use provided time or create synthetic time if not given
+      if time is None:
```
marcomuzio, line 49: I think we can remove this block.

araproc/analysis/snr.py
```diff
@@ -1,26 +1,41 @@
...
     waveform: TGraph or np.ndarray
         A TGraph or np.ndarray of the waveform voltage.
+    order : int, optional (Default = 1)
+        How many points on each side to compare for finding extrema.
+    use_local: bool, optional (Default = True)
+        If True, returns the peak-to-peak voltage based on local extrema. 
+        If False, returns the peak-to-peak voltage based on global extrema.
+    time_window: float, optional (Default = 20 ns)
+        Time window in nanoseconds to calculate the peak-to-peak voltage.
+    dt : float, optional (Default = 0.5 ns)
```
marcomuzio, line 23: We can drop the last two arguments.

araproc/analysis/snr.py
```diff
@@ -1,26 +1,41 @@
...
 
-    """
-    Calculates the peak-to-peak voltage of a voltage trace.
+    """ 
+    Calculates the peak-to-peak voltage of a signal using local (Default) and global extrema.
 
     Parameters
     ----------
     waveform: TGraph or np.ndarray
         A TGraph or np.ndarray of the waveform voltage.
```
marcomuzio, line 15: Let's update this to reflect that the user now needs to pass a TGraph, tuple, or list.

marcomuzio (changes requested):
Sorry I was slow to turn this around. Let's make this last change and then we should be good to go. Thanks!

Mohammad-Neutrino:
Hi @marcomuzio , I made the final suggested changes. Let me know how it looks now. Thanks!

marcomuzio:
@Mohammad-Neutrino I adjusted this a bit to try to speed this up a bit. I think it should still essentially do the same thing, but could you verify you think it's still working as intended?

Mohammad-Neutrino:
Hi @marcomuzio, this is as fast as the original VPP and indeed gives the desired output for pkpk amplitude in local extrema within 20 ns. Thank you for the speed ups!! I think it's good to go now?

marcomuzio:
Yes, let's merge!

Mohammad-Neutrino:
Merging now! Thanks again @marcomuzio!

PR #102, opened 2025-03-28
Fixed if statement indent level.

marcomuzio:
`if` statement to throw an error if no filter covering a flagged frequency is found was _inside_ the `for` loop over filters, which would cause a crash if the first filter checked did not cover that frequency. Fixed so that the check is done _after_ looping over all filters.

### DIFF
--- a/araproc/analysis/cw_filter.py
+++ b/araproc/analysis/cw_filter.py
@@ -138,8 +138,8 @@ def get_active_filters(cw_filters, cw_ids, chan):
 
                     active_filters[filter_i] = filter
 
-                if not isFiltered:
-                    raise Exception(f"IDed CW at {freq} GHz has no corresponding filter! Please add one and rerun.")
+            if not isFiltered:
+                raise Exception(f"IDed CW at {freq} GHz has no corresponding filter! Please add one and rerun.")
 
     # filters are applied in order they appear in dict
     # there's some advantage to apply them in order of descending 
--- a/araproc/analysis/cw_filter.py
+++ b/araproc/analysis/cw_filter.py
@@ -73,9 +73,14 @@ def apply_filters(cw_filters, waveform_bundle, cw_ids = None):
     check_cw_ids(cw_ids)
 
     filtered_waveforms = {}
+    active_cw_filters_v = get_active_filters(cw_filters, cw_ids, 0)
+    active_cw_filters_h = get_active_filters(cw_filters, cw_ids, 8)
     for ch_id, wave in waveform_bundle.items():
 
-        active_cw_filters = get_active_filters(cw_filters, cw_ids, ch_id)
+        if ch_id in const.vpol_channel_ids:
+            active_cw_filters = active_cw_filters_v
+        else:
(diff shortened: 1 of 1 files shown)


araproc/analysis/cw_filter.py
```diff
@@ -73,9 +73,14 @@ def apply_filters(cw_filters, waveform_bundle, cw_ids = None):
     check_cw_ids(cw_ids)
 
     filtered_waveforms = {}
+    active_cw_filters_v = get_active_filters(cw_filters, cw_ids, 0)
+    active_cw_filters_h = get_active_filters(cw_filters, cw_ids, 8)
     for ch_id, wave in waveform_bundle.items():
 
-        active_cw_filters = get_active_filters(cw_filters, cw_ids, ch_id)
+        if ch_id in const.vpol_channel_ids:
```
clark2668, line 80: I guess I don't understand why you have to recompute this every time?

araproc/analysis/cw_filter.py
```diff
@@ -118,28 +123,25 @@ def get_active_filters(cw_filters, cw_ids, chan):
...
+            mask = np.logical_and(badFreqs >= fmin, badFreqs <= fmax) # find frequencies covered by this filter  
+            isFiltered = np.logical_or(isFiltered, mask) # track frequencies that have covered by a filter
+ 
+            # if filter covers any flagged frequency, activate it 
+            if mask.any(): 
+                active_filters[filter_i] = filter
 
-                if not isFiltered:
-                    raise Exception(f"IDed CW at {freq} GHz has no corresponding filter! Please add one and rerun.")
+        if not isFiltered.all():
```
clark2668, line 142: You need this because isFiltered is an array? Just confirming my understanding

clark2668 (commented):
Left a comment about something I'm confused about.

araproc/analysis/cw_filter.py
marcomuzio: Sorry, what do you mean? This if statement was added to avoid recomputing each time. If I did it right, it computes once per polarization.

araproc/analysis/cw_filter.py
marcomuzio: Right `isFiltered` is an array of booleans tracking whether or not a filter was found for each flagged frequency.

clark2668: Right, gotcha. Then okay.

araproc/analysis/cw_filter.py
clark2668: No no, it does. I guess this is fine. Seems a bit of a shame you have to call `get_active_filters` every event, but maybe that's required by the fact that which filters you want turned on changes per event?

marcomuzio: Right ya, I don't think there's a way around that unfortunately.

marcomuzio: All good here?

clark2668: yes, good to go!

PR #103, opened 2025-04-01
Added fast filter variables

marcomuzio:
This PR adds new analysis variables intended for use a low-cost rejection of the bulk of thermal events. These include an average spacetime interval and a (slightly modified/simplified) wavefront RMS.

### DIFF
--- a/araproc/analysis/fast_filter.py
+++ b/araproc/analysis/fast_filter.py
@@ -3,6 +3,7 @@ import ROOT
 
 from araproc.framework import waveform_utilities as wfu
 from araproc.framework import constants as const
+from araproc.analysis import snr
 
 def get_peak_time(waveform):
 
@@ -37,8 +38,7 @@ def get_peak_time(waveform):
     else:
       raise Exception("Unsupported data type in fast_filter.get_peak_time. Abort")
 
-    idxPeak = np.argmax(np.abs(trace))
-    tPeak = time[idxPeak]
+    tPeak, _ = snr.get_windowed_vpp(time, trace)
 
     return tPeak
 
--- a/araproc/analysis/fast_filter.py
+++ b/araproc/analysis/fast_filter.py
@@ -1,0 +1,173 @@
+import numpy as np
+import ROOT
+
+from araproc.framework import waveform_utilities as wfu
+from araproc.framework import constants as const
+
+def get_peak_time(waveform):
+
+    """
+    Finds the time of the absolute maximum of the trace.
+
+    Parameters
+    ----------
+    waveform: ROOT.TGraph, tuple, or list
+        A TGraph or a tuple/list containing two np.ndarrays: (time, trace).
+  
+    Returns
+    -------
+    tPeak : float
+        Time of the peak.
+    """
+
+    if(isinstance(waveform, ROOT.TGraph)):
+      time, trace = wfu.tgraph_to_arrays(waveform)
+      if len(trace) != len(time):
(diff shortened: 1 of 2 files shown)


araproc/analysis/fast_filter.py
```diff
@@ -0,0 +1,173 @@
...
+        raise ValueError("Trace and time must have the same length.")
+
+    else:
+      raise Exception("Unsupported data type in fast_filter.get_peak_time. Abort")
+
+    tPeak, _ = snr.get_windowed_vpp(time, trace)
+
+    return tPeak
+
+def get_wavefront_rms(wave_bundle, station_id, excluded_channels=[]):
```
clark2668, line 45: It occurs to me that this method will allow things to be cross polarized. Or is the idea that that's fine, and it's up to the user to call this twice, once with v exclusion, and once with h exclusion? But the code is generic?

araproc/analysis/fast_filter.py
```diff
@@ -0,0 +1,173 @@
...
+
+    # calculate rms
+    wavefront_rms = np.mean(cosTheta**2) - np.mean(cosTheta)**2
+
+    return wavefront_rms
+
+def get_avg_spacetime_interval(wave_bundle, station_id, excluded_channels=[]):
+
+    """
+    Calculates the average spacetime interval.
```
clark2668, line 115: More docs please. What is the spacetime interval?

clark2668 (changes requested):
This would benefit from a use case. Please show how to use this in e.g. `example.py`.

araproc/analysis/fast_filter.py
marcomuzio: Right, in principle it would allow that and it's up to the user. I think that's the case for a lot of our functions in AraProc. Maybe we get away with this becuz the implementation here was lazier than what you did for A2/3 (eg we don't actually form baseline pairs or find tetrads etc).

clark2668: Right, okay. Is doing the pair creation fast? It occurs to me that if it's slow, you could make this code faster by passing that in, rather than having the code remake them constantly.

marcomuzio: It doesn't really generate the pairs, it just does a nested for loop. But I think the number of iterations is the same as if you passed the pairs ahead of time. Maybe I am wrong.

araproc/analysis/fast_filter.py
marcomuzio: Updated, let me know what you think.

araproc/analysis/fast_filter.py
clark2668: Okay, fair enough. So long as we haven't increased the runtime (substantially) by not precomputing the pairs. Although, I'd point out that the RTC has a "pair creation" utility available. So there is a bit of duplicated code here for doing the combinatorics. But okay, I won't push this, it's nice that this is self contained :)

araproc/analysis/fast_filter.py
clark2668: Can you add one more line about what you assume for v (I guess it's c/n_avg?), and then I think this is good to go!

araproc/analysis/fast_filter.py
marcomuzio: Ya I don't think it adds any significant overhead. And fair enough!

marcomuzio:
Thanks! Merging now!

PR #104, opened 2025-04-05
Added run number parser

marcomuzio:
Added helper function to parse the run number from a ROOT file directly.

### DIFF
--- a/araproc/framework/file_utilities.py
+++ b/araproc/framework/file_utilities.py
@@ -94,3 +94,32 @@ def guess_location_of_daq_config_file(station_id, run_number):
     config_file = files[0]
 
     return config_file
+
+def parse_run_number(run_file_path):
+  """
+  Helper function to parse the run number from a run file path.
+  Parameters
+  ----------
+  run_file_path : str
+      Path to the run file.
+  
+  Returns
+  -------
+  run_number : int
+      Run number of the run file.
+  """
+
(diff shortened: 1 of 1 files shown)


clark2668:
Looks good -- are you going to update the DataWrapper correspondingly? Seems like that could go in at once?

marcomuzio:
I could, I guess it's up to you. My gut was to leave it as is, where it gets the run from the file's data itself? That seems more foolproof, this is kind of a cheat. What do you think?

marcomuzio:
I guess we could add it as like a way to catch mislabelled file names? But idk if thats what we want here necessarily

clark2668:
Oh no, you’re right. I forgot that I pulled it from the file directly rather than the file name. So we’re good. Continue as planned! :)This message was sent from a mobile device, please excuse any brevity or errors.On Apr 5, 2025, at 6:05 PM, Marco Muzio ***@***.***> wrote:﻿
  I guess we could add it as like a way to catch mislabelled file names? But idk if thats what we want here necessarily—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because your review was requested.Message ID: ***@***.***>



marcomuzio left a comment (ara-software/AraProc#104)
I guess we could add it as like a way to catch mislabelled file names? But idk if thats what we want here necessarily

PR #105, opened 2025-04-08
Make ray tracing tables an environment variable

abigailbishop:
Since the Ray Trace timing tables need to be flexible I made the path to the ray tracing tables yet another environment variable required for the analysis.

### DIFF
--- a/araproc/analysis/__init__.py
+++ b/araproc/analysis/__init__.py
@@ -7,7 +7,7 @@ ray_trace_tables_dir = os.getenv('RAY_TRACE_TABLES', None)
 
 # If the environment variable is not set, use a default
 if not ray_trace_tables_dir:
-    ray_trace_tables_dir = "/cvmfs/icecube.osgstorage.org/icecube/PUBLIC/groups/arasoft/raytrace_timing_tables/"
+    ray_trace_tables_dir = "/data/ana/ARA/processing/support/raytrace_timing_tables/"
     warnings.warn(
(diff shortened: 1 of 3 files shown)


abigailbishop:
@marcomuzio Did you want to apply the "if RAY_TRACE_TABLES isn't set, use a default but bring a warning" setup here as well?

marcomuzio:
Sorry, ya let's add the default since I think this is a pretty specific use case. Most users won't need to think about it (either cuz they are on cobalt or will inherit an environment from you for the grid).

abigailbishop:
Done!

araproc/analysis/__init__.py
```diff
@@ -0,0 +1,23 @@
+import os
+import warnings
+
+# Get the directory of ray tracing tables from an environment variable,
+#  with a fallback or error if not set
+ray_trace_tables_dir = os.getenv('RAY_TRACE_TABLES', None)
+
+# If the environment variable is not set, use a default
+if not ray_trace_tables_dir:
+    ray_trace_tables_dir = "/cvmfs/icecube.osgstorage.org/icecube/PUBLIC/groups/arasoft/raytrace_timing_tables/"
```
marcomuzio, line 10: Sorry can we use `/data/ana/ARA/processing/support/raytrace_timing_tables` as default? I think that was the more stable one

marcomuzio (approved):
Just one small change then good to go!

araproc/analysis/__init__.py
abigailbishop: Yep done.

PR #107, opened 2025-04-09
Update A3 string masking and add some filters to A2

abigailbishop:
Add a 299 MHz filter to some configs on A2 it's been noticed on. Remove a string from A3 configs 7, 8, 9.


marcomuzio (approved):
Looks good!

PR #109, opened 2025-04-14
Add CW ID frequency threshold and trace types for cw filtering- and bandpassing-only

marcomuzio:
This PR adds a minimum CW ID frequency above which CW ID will be used to turn on filters. Below this frequency, all filters are turned on. (Ideally this would be removed someday.) The default is 120 MHz, which is what we used in our phase variance processing, but it can be adjusted by passing an argument when making the dataset object.

Also added two trace types: `cw_filtered` and `bandpassed` which only apply their respective filter.

### DIFF
--- a/araproc/analysis/cw_filter.py
+++ b/araproc/analysis/cw_filter.py
@@ -40,7 +40,7 @@ def apply_filters_one_channel(cw_filters, waveform_in):
     
     return latest_waveform
 
-def apply_filters(cw_filters, waveform_bundle, cw_ids = None):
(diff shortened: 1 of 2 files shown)


clark2668:
So, I think this mostly looks okay. I just want to be sure tho: if people ask for CW filtered traces, they still get filtered before being bandpassed, yes?

marcomuzio:
Right the order of operations hasn't changed. "Filtered" still return waveforms that are cw filtered and then bandpassed.

clark2668:
Okay, then I think this is good to go.

PR #111, opened 2025-04-16
Ensure CW filters are always sorted

marcomuzio:
This PR ensures CW filters are always applied in order of descending `min_power_ratio`. This was already enforced when CW ID is being used and previously was enforced by users via the ordering in yaml files when CW ID wasn't used. This update ensures the order that CW filters are applied is uniform.

Also added a getter function to have easier access to the flagged CW frequencies in each polarization of an event.

### DIFF
--- a/araproc/framework/dataset.py
+++ b/araproc/framework/dataset.py
@@ -457,12 +457,26 @@ class DataWrapper:
         ROOT.SetOwnership(useful_event, True)
         
         # if we have cw ids loaded, get the corresponding entry to this event 
-        if self.cw_id_tree is not None:
-            event_number = useful_event.eventNumber
-            if self.cw_id_tree.GetEntryWithIndex(event_number) < 0:
-                raise Exception(f"Unable to get corresponding cw id entry for {event_number}.")
+        event_number = useful_event.eventNumber
+        self.get_cw_id_entry(event_number)
  
         return useful_event
+
+    def get_cw_id_entry(self, event_number):
+        """
+        Get the CW ID tree entry for an event.
+
+        Parameters
+        ----------
+        event_number : int
+          Event number to load CW ID for.
+        """
+
+        if self.cw_id_tree is not None:
+            if self.cw_id_tree.GetEntryWithIndex(event_number) < 0:
+                raise Exception(f"Unable to get corresponding cw id entry for {event_number}.")
+        
+        return 
 
     def get_cw_ids(self):
         """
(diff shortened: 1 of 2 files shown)


araproc/framework/dataset.py
```diff
@@ -475,6 +489,46 @@ def get_cw_ids(self):
...
+        event_number = useful_event.eventNumber
+        self.get_cw_id_entry(event_number)
+  
+        cw = {"v" : [],
+              "h" : []}
+        cw_ids = self.get_cw_ids()
+        if cw_ids is None:
+             raise Exception("No event loaded, so I can't tell you what the CW frequencies are!")
+
+        scan_directions = ["fwd", "bwd"]
```
abigailbishop, line 519: Can we get a comment explaining this block since not everyone might infer that "fwd" and "bwd" mean forwards and backwards?

araproc/framework/dataset.py
```diff
@@ -475,6 +489,46 @@ def get_cw_ids(self):
...
+        cw_v, cw_h : numpy arrays
+            Arrays of the IDed frequencies in both scan directions for each polarization.
+
+        """
+        
+        # if we have cw ids loaded, get the corresponding entry to this event 
+        event_number = useful_event.eventNumber
+        self.get_cw_id_entry(event_number)
+  
+        cw = {"v" : [],
```
abigailbishop, line 513: Supremely nitpick but `cw` isn't used until the next coding block and could be moved there. Then the comment preceding the next code block can contain a hint that CW filters are being loaded in according to scans of VPols and HPols in the forwards and backwards directions.

abigailbishop:
This looks good Marco! I left a couple little comments but I think this is good to go regardless of if you act on the comments or not.

araproc/framework/dataset.py
marcomuzio: Done!

araproc/framework/dataset.py
```diff
@@ -475,6 +489,46 @@ def get_cw_ids(self):
...
+              "h" : []}
+        cw_ids = self.get_cw_ids()
+        if cw_ids is None:
+             raise Exception("No event loaded, so I can't tell you what the CW frequencies are!")
```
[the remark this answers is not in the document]
marcomuzio: Done!

araproc/framework/dataset.py
marcomuzio: Done, let me know what you think

marcomuzio:
Gonna go ahead and merge!

PR #113, opened 2025-04-18
CW ID leak fix

marcomuzio:
This PR substantially reduces memory leaks due to CWID. These seem to originate from the way the TTree is read in and navigated, which somehow causes ROOT to allocate memory in the background. The TTreeReader avoids this. Remaining memory leaks might similarly be related to how TTrees are being read in.

Some changes we also introduced to the cw_filter module to avoid creating dictionaries of SineSubtract filters. I'm not sure if these were actually a problem, so we could revert that part of the code as the old version was a bit simpler in my opinion.

Also removed the previously added getter function `get_event_cw` as it's now redundant.

### DIFF
--- a/araproc/analysis/cw_filter.py
+++ b/araproc/analysis/cw_filter.py
@@ -6,7 +6,7 @@ import numpy as np
 
 from araproc.framework import constants as const
 
-def apply_filters_one_channel(cw_filters, waveform_in):
+def apply_filters_one_channel(cw_filters, active_cw_filters, waveform_in):
     """
     Apply CW filters to a single waveform.
     This function will not delete `waveform` for you,
@@ -19,6 +19,8 @@ def apply_filters_one_channel(cw_filters, waveform_in):
         A dictionary of sine subtract filters to be applied.
         Key is a string, and is the name of the filter from the config file.
         Value is the sine subtract filter to applied.
+    active_cw_filters : dictionary
+        A dictionary of booleans tracking which filters are activated.
     waveform_in : ROOT.TGraph
         A single waveform to be filtered.
     
@@ -32,8 +34,10 @@ def apply_filters_one_channel(cw_filters, waveform_in):
     ROOT.SetOwnership(latest_waveform, True) # take posession
 
     if len(cw_filters) > 0:
-        for filter_i, filter in cw_filters.items():
-            local_waveform = filter["filter"].subtractCW(latest_waveform, -1)
+        for filter_i, isActive in active_cw_filters.items():
+            if not isActive:
+                continue
+            local_waveform = cw_filters[filter_i]["filter"].subtractCW(latest_waveform, -1)
             ROOT.SetOwnership(local_waveform, True) # take posession 
             del latest_waveform
             latest_waveform = local_waveform # update the latest waveform
@@ -56,8 +60,8 @@ def apply_filters(cw_filters, waveform_bundle, cw_ids = None, min_cw_id_freq = 0
         A dictionary of waveforms to be processed.
         Key is channel id.
         Value is the TGraph to be filtered.
-    cw_ids : dictionary
-        A dictionary of cw id info.
+    cw_ids : tuple
+        A tuple of numpy arrays containing cw id info.
         If None, all set CW filters allowed to filter. Otherwise, only those 
         covering frequencies in the cw id info will be allowed to filter.
 
@@ -81,7 +85,7 @@ def apply_filters(cw_filters, waveform_bundle, cw_ids = None, min_cw_id_freq = 0
             active_cw_filters = active_cw_filters_v
         else:
             active_cw_filters = active_cw_filters_h
-        filtered_waveforms[ch_id] = apply_filters_one_channel(active_cw_filters, wave)
+        filtered_waveforms[ch_id] = apply_filters_one_channel(cw_filters, active_cw_filters, wave)
 
 
     return filtered_waveforms
@@ -96,8 +100,8 @@ def get_active_filters(cw_filters, cw_ids, chan, min_cw_id_freq):
         A dictionary of sine subtract filters available to be applied..
         Key is a string, and is the name of the filter from the config file.
         Value is the sine subtract filter to applied.
-    cw_ids : dictionary
-        A dictionary of cw id info.
+    cw_ids : tuple
+        A tuple of numpy arrays containing cw id info.
         If None, all available CW filters are activated. Otherwise, only those 
         covering frequencies in the cw id info will be activated.
     chan : int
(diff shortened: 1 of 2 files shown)


araproc/analysis/cw_filter.py
```diff
@@ -106,54 +110,47 @@ def get_active_filters(cw_filters, cw_ids, chan, min_cw_id_freq):
...
     for filter_i, filter in cw_filters.items():
         fmax = filter["max_freq"]
 
         # if filter covers region below cw id threshold, activate it 
         if fmax < min_cw_id_freq:
-            active_filters[filter_i] = filter
+            active_filters[filter_i] = True
             
     
     if chan in const.vpol_channel_ids:
```
abigailbishop, line 133: Can we get a comment here explaining the switching of the `_` and `badFreqs` variables?

marcomuzio: Done, how's that?

abigailbishop: Yeah, it still took me a minute to understand what was happening but it makes more sense now, thanks!

araproc/analysis/cw_filter.py
```diff
@@ -6,7 +6,7 @@
 
 from araproc.framework import constants as const
 
-def apply_filters_one_channel(cw_filters, waveform_in):
+def apply_filters_one_channel(cw_filters, active_cw_filters, waveform_in):
     """
     Apply CW filters to a single waveform.
     This function will not delete `waveform` for you,
```
abigailbishop, line 12: There's no object named `waveform` anymore. Which object is this warning referring to?

araproc/analysis/cw_filter.py
```diff
@@ -96,8 +100,8 @@ def get_active_filters(cw_filters, cw_ids, chan, min_cw_id_freq):
         A dictionary of sine subtract filters available to be applied..
         Key is a string, and is the name of the filter from the config file.
         Value is the sine subtract filter to applied.
```
abigailbishop, line 102: Typo. "filter to be applied." or "filter to apply."

araproc/analysis/cw_filter.py
marcomuzio: Ya sorry, I hate how this portion got more complicated... maybe I should revert it? I'm not sure it's doing anything for us tbh.

araproc/analysis/cw_filter.py
```diff
@@ -106,54 +110,48 @@ def get_active_filters(cw_filters, cw_ids, chan, min_cw_id_freq):
...
-            fmax = filter["max_freq"]
-            
-            mask = np.logical_and(badFreqs >= fmin, badFreqs <= fmax) # find frequencies covered by this filter  
-            isFiltered = np.logical_or(isFiltered, mask) # track frequencies that have covered by a filter
- 
-            # if filter covers any flagged frequency, activate it 
-            if mask.any(): 
-                active_filters[filter_i] = filter
+        mask = np.logical_and(badFreqs >= fmin, badFreqs <= fmax) # find frequencies covered by this filter  
+        isFiltered = np.logical_or(isFiltered, mask) # track frequencies that have covered by a filter
```
abigailbishop, line 146: Typo, I think you want "track frequencies that have been covered by a filter"

araproc/analysis/cw_filter.py
marcomuzio: I'm... not sure. I think it's been this way for a while. I'd guess it's referring to `local_waveform`? Maybe @clark2668 can shed some light

araproc/analysis/cw_filter.py
```diff
@@ -106,54 +110,48 @@ def get_active_filters(cw_filters, cw_ids, chan, min_cw_id_freq):
...
-            fmin = filter["min_freq"]
-            fmax = filter["max_freq"]
-            
-            mask = np.logical_and(badFreqs >= fmin, badFreqs <= fmax) # find frequencies covered by this filter  
-            isFiltered = np.logical_or(isFiltered, mask) # track frequencies that have covered by a filter
- 
-            # if filter covers any flagged frequency, activate it 
-            if mask.any(): 
-                active_filters[filter_i] = filter
+        mask = np.logical_and(badFreqs >= fmin, badFreqs <= fmax) # find frequencies covered by this filter
```
abigailbishop, line 145: Can this comment be rephrased slightly to something like "Array of booleans reflecting which of the frequencies in`badFreqs` are covered by this filter" 
the variable name also took me a moment to parse so maybe like `badFreqs_covered` or something would be more intuitive?

araproc/framework/dataset.py
```diff
@@ -96,6 +96,8 @@ class DataWrapper:
         the pointer to the ROOT TFile that corresponds to the opened cw id file
     cw_id_tree : ROOT TTree
         the sound the the mousetrap makes
```
abigailbishop, line 98: What is this comment?

araproc/framework/dataset.py
```diff
@@ -114,8 +116,8 @@ class DataWrapper:
         number of data events in the data ROOT file
     raw_event_ptr : RawAtriStationEvent
         a RawAtriStationEvent ptr from AraRoot
-    cw_id_ptrs : dict of ptrs
-        ptrs to Branches of the cw id ttree
+    cw_id_reader_vals : dict
+        dictionary of reader values for different branches
```
abigailbishop, line 120: Can you be more specific here? This is TTreeReader values or pointers?

abigailbishop: Oh I see they're TTreeReaderValue objects. Can you state this explicitly?

araproc/framework/dataset.py
```diff
@@ -455,83 +451,58 @@ def get_useful_event(self, event_idx):
...
+        # reset reader -- reader MUST move sequentially thru TTree
+        self.cw_id_reader.Restart()
         
-        # if we have cw ids loaded, get the corresponding entry to this event 
-        event_number = useful_event.eventNumber
-        self.get_cw_id_entry(event_number)
+        # iterate thru reader until we find the right entry
+        # we need to IMMEDIATELY readout the reader data, otherwise there's
+        # no guarantee the reader will point to the right (or any) entry
+        while self.cw_id_reader.Next():
```
abigailbishop, line 485: Clarification question, so for every event the TTreeReader needs to reset to the first entry and then iterate through every entry until the requested event is opened? Is that going to take a concerning amount of more time at the end of processing compared to the beginning or is this fine?

araproc/analysis/cw_filter.py
abigailbishop: I think I just don't understand why you can't call `cw_ids[0]` and `cw_ids[1]`. Is there some benefit to `_, bad_freqs = cw_ids`?

marcomuzio: I think either work? Should be equivalent.

araproc/framework/dataset.py
abigailbishop: There's another comment like this on line 94

marcomuzio: Right, this is line 94's partner.

araproc/analysis/cw_filter.py
abigailbishop: Then I think cw_ids[0] is more readable but alas I don't know how many people will be reading this code.

araproc/framework/dataset.py
marcomuzio: Are you asking for it to be `cw_id_reader_vals : dict of TTreeReader values`?

araproc/framework/dataset.py
marcomuzio: Ya that's right. Unfortunately, TTreeReaders cannot randomly access entries in the TTree, they _have_ to read thru the TTree sequentially. I was also worried about this from a performance perspective, but running L2 it didn't seem to create any noticeable slow down afaict.

araproc/framework/dataset.py
abigailbishop: Yeah that would help. I just didn't know what "reader values" meant so having the term TTreeReaderValues listed somewhere in this comment helped me understand what was happening a bit better

araproc/framework/dataset.py
abigailbishop: Sounds good. I'll keep an eye on the run times for the 10pct and 100pct L2 I'm running right now incase this does make them take longer

marcomuzio: Sounds great! Fingers crossed it's not an issue cuz idk what we'll do ugh

araproc/analysis/cw_filter.py
marcomuzio: Haha true. But changed it!

araproc/analysis/cw_filter.py
marcomuzio: Done.

araproc/analysis/cw_filter.py
marcomuzio: Done. I should not be writing comments so late at night lmao.

araproc/framework/dataset.py
marcomuzio: Done.

araproc/framework/dataset.py
marcomuzio: Well at the very least, it's still faster than when I turn CWID off.

araproc/analysis/cw_filter.py
abigailbishop: I don't think it's desperately urgent. We can merge this and come back to this later if we want

PR #114, opened 2025-04-18
Ensure that bandpassed trace is properly assigned.

marcomuzio:
Fixed small bug in the assignment of that bandpassed-only waveform.


abigailbishop:
Looks good to me!

PR #115, opened 2025-04-18
Fixed bug in windowed vpp

marcomuzio:
Adjusted how we implemented the computationally-efficient version of the search for Vpp in a time window. This should now essentially never fail to calculate a Vpp.

### DIFF
--- a/araproc/analysis/snr.py
+++ b/araproc/analysis/snr.py
@@ -29,6 +29,10 @@ def get_windowed_vpp(time, trace, time_window = 20):
     if trace.max() == trace.min():
       return 0.0, 0.0
 
+    # to speed up computation, we only consider points beyond some threshold
+    # initially, this threshold is set to 0.5 of the max/min value, which nearly 
+    # always is sufficient to find the right Vpp, but in the rare case it's not
+    # we lower the threshold and try again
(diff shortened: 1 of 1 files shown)


araproc/analysis/snr.py
```diff
@@ -29,29 +29,43 @@ def get_windowed_vpp(time, trace, time_window = 20):
...
-
-    # limit to extrema within time_window of each other
-    mask = np.abs(all_dt) <= time_window
-    if not mask.any():
-      return 0.0
-    
+    # to speed up computation, we only consider points beyond some threshold
+    # initially, this threshold is set to 0.5 of the max/min value, which nearly 
+    # always is sufficient to find the right Vpp, but in the rare case it's not
+    # we lower the threshold and try again
```
abigailbishop, line 35: Can you say "lower the threshold by half and try again"

abigailbishop:
Only one small comment- looks great!

araproc/analysis/snr.py
marcomuzio: Done!

abigailbishop: Thanks, merge away!

PR #116, opened 2025-04-21
Fixed expected readout lengths for A3 config 1

marcomuzio:
Adjusted the expected readout lengths for A3 config 1 to the right values from data. These were set incorrectly due to this number being missing from most of the DAQ config logs that the scraping script pulls data from. The result was that essentially every event in the config was flagged as bad quality, even though there is nothing outwardly wrong with it. This fixes that error.


Mohammad-Neutrino:
Did the soft trigger length also check out to be 160 ns in data?

marcomuzio:
Ah good point, I forgot to check before. I just spot checked it and indeed the software trigger readout length looks to be 8 blocks! So all should be okay.

Mohammad-Neutrino:
Great, then I think it's good to go! 🚀

PR #121, opened 2025-04-25
Add event vertex and direction to simulation information object

abigailbishop:
The direction saved is the one directly from AraSim. 

The XY position is a rough translation from AraSim and may not preserve the ice flow coordinates. @AlanSalcedo I'd love your input on this if you have the time. I wasn't sure if station.GetX() - event.GetX() was a valid calculation. 

The depth uses the FlatToEarth_SameAngle conversion from the AraSim retracer.

### DIFF
--- a/araproc/framework/dataset.py
+++ b/araproc/framework/dataset.py
@@ -855,12 +855,16 @@ class SimWrapper:
             s, a = trig_ants[ant_idx]
 
             # Get the SNR of this antenna's waveform
-            waveform = wu.arrays_to_tgraph(
-                np.asarray(self.report_ptr.stations[0].strings[s].antennas[a].time_mimic), 
-                np.asarray(self.report_ptr.stations[0].strings[s].antennas[a].V_mimic),
-            )
-            SNR = get_snr(waveform)
+            t = self.report_ptr.stations[0].strings[s].antennas[a].time_mimic    
+            ROOT.SetOwnership(t, False) # give python full ownership (see README.md for more info)
+            v = self.report_ptr.stations[0].strings[s].antennas[a].V_mimic
+            ROOT.SetOwnership(v, False)
+            waveform = ROOT.TGraph(len(t), t.data(), v.data())
+            ROOT.SetOwnership(waveform, True)
 
+            #waveform = wu.arrays_to_tgraph(np.linspace(0, 1, 100), np.random.normal(size=100))
+            SNR = get_snr(waveform)
+  
             # If this antenna's waveform is greater than the saved SNR, save this 
             #   antennas string index, antenna index, and SNR as the best
             if SNR > best_SNR: 
--- a/araproc/framework/dataset.py
+++ b/araproc/framework/dataset.py
@@ -862,7 +862,6 @@ class SimWrapper:
             waveform = ROOT.TGraph(len(t), t.data(), v.data())
             ROOT.SetOwnership(waveform, True) # python's responsibility
 
-            #waveform = wu.arrays_to_tgraph(np.linspace(0, 1, 100), np.random.normal(size=100))
             SNR = get_snr(waveform)
   
             # If this antenna's waveform is greater than the saved SNR, save this 
--- a/araproc/framework/dataset.py
+++ b/araproc/framework/dataset.py
@@ -776,8 +776,8 @@ class SimWrapper:
         sim_info["pid"] = self.event_ptr.nu_prim_pid
         sim_info["evid"] = self.event_ptr.event_ID
         sim_info["vertex"] = self.get_AraSim_xyz_position(
-            self.event_ptr.Nu_Interaction[likely_interaction].posnu, 
-            self.detector_ptr.stations[0])
+            self.detector_ptr.stations[0],
+            self.event_ptr.Nu_Interaction[likely_interaction].posnu)
         sim_info["direction"] = (
             self.event_ptr.Nu_Interaction[likely_interaction].nnu.Theta(),
             self.event_ptr.Nu_Interaction[likely_interaction].nnu.Phi()
@@ -879,10 +879,10 @@ class SimWrapper:
         ----------
         origin : AraSim::Position
             Location to measure X-Y displacement from `position` and depth from ice surface.
-            Usually a cascade.
+            Usually the station.
         position : AraSim::Position
             Location to measure `origin` X-Y displacement with respect to. 
-            Usually the station.
+            Usually a cascade.
 
         Returns
         -------
@@ -893,11 +893,6 @@ class SimWrapper:
         depth : float
             Depth of `origin` with respect to the surface of the ice defined by `ROOT::IceModel`
         """
-        # Convert XY position coordinates
-        r_from_pole_position = np.sqrt(position.GetX()**2 + position.GetY()**2)
-        lon_position = np.radians(position.Lon())
-        x_position = r_from_pole_position * np.cos(lon_position)
-        y_position = r_from_pole_position * np.sin(lon_position)
 
         # Get XY coordinates of provided origin
         r_from_pole_origin = np.sqrt(origin.GetX()**2 + origin.GetY()**2)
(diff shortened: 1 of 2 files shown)


araproc/framework/dataset.py
```diff
@@ -73,6 +74,142 @@ def get_filters(station_id, analysis_config):
...
+    return best_ant
+
+def get_AraSim_xyz_position(source, target, icemodel):
+    """
+    Return the XY displacement of a source and its depth with respect to
+      the surface of the ice. 
+
+    Parameters
+    ----------
+    target : AraSim::Position
```
marcomuzio, line 175: Can we call these `origin` and `vertex`? Somehow these names are confusing to me.

araproc/framework/dataset.py
```diff
@@ -73,6 +74,142 @@ def get_filters(station_id, analysis_config):
...
+        s, a = trig_ants[ant_idx]
+
+        # Get the SNR of this antenna's waveform
+        waveform = [
+            np.asarray(file.AraTree2.report.stations[0].strings[s].antennas[a].time_mimic), 
+            np.asarray(file.AraTree2.report.stations[0].strings[s].antennas[a].V_mimic),
+        ]
+        vpp = get_vpp(waveform)
+        rms = get_jackknife_rms(waveform[1])
+        SNR = vpp/rms/2.0
```
marcomuzio, line 157: Why not use `snr.get_snr` directly?

araproc/framework/dataset.py
```diff
@@ -73,6 +74,142 @@ def get_filters(station_id, analysis_config):
...
+    best_ant : tuple
+        Tuple containing the index of this antenna in the 
+        `file.AraTree2.report.stations[0].strings` and the 
+        `file.AraTree2.report.stations[0].strings[string_index].antennas`
+        objects, respectively
+    """
+
+    # Alias the station we're analyzing as it is in the report and detector class
+    station_r = file.AraTree2.report.stations[0]
+    station_d = file.AraTree.detector.stations[0]
```
marcomuzio, line 99: Have you run this over a bunch of events and tracked the memory? These ROOT objects often need to be handled really carefully to avoid leaks. But maybe it's fine here, it's always hard to tell whats gonna set pyroot off.

marcomuzio (changes requested):
Looks good to me. Just a few small things.

araproc/framework/dataset.py
abigailbishop: Because of a little bug in get_snr and I was too lazy to resolve that I maybe just could. Essentially get_vpp isn't set up to process the waveform from get_snr if it isn't a TGraph even though we advertise that a numpy array can be used in get_snr. 

Here's the handling of the waveform in get_snr: 
```
    vpp = get_vpp(waveform)
    rms = get_jackknife_rms(waveform)
```
And here's how it was handled in this script: 
```
        vpp = get_vpp(waveform)
        rms = get_jackknife_rms(waveform[1])
```

araproc/framework/dataset.py
abigailbishop: I ran the A2 noise sim L2 generation and over 500 events the memory usage went from 965 MB to 1423 MB

araproc/framework/dataset.py
```diff
@@ -756,11 +894,26 @@ def get_sim_information(self,
...
+
         sim_info = {}
         sim_info["weight"] = self.event_ptr.Nu_Interaction[0].weight
         sim_info["enu"] = self.event_ptr.nu_prim_energy
+        sim_info["pid"] = self.event_ptr.nu_prim_pid
         sim_info["evid"] = self.event_ptr.event_ID
-        sim_info["pid"] = self.event_ptr.prim_particle_type
+        sim_info["vertex"] = get_AraSim_xyz_position(
+            self.event_ptr.Nu_Interaction[likely_interaction].posnu, 
+            self.root_tfile.AraTree.detector.stations[0],
```
AlanSalcedo, line 910: I checked these in AraSim and posnu is a Position.cc object that is a Vector. In the Detector class, stations is alsoa vector. GetX(), GetY(), GetZ() should work on both. I'm just curious what station[0].GetX() -or Y and Z- contain. Hopefully the station center?

AlanSalcedo: Okay, I misunderstood your request. Idk if the subtraction would work but I suppose you already tested it?

araproc/framework/dataset.py
marcomuzio: ugh wut, i thought we fixed that?

araproc/framework/dataset.py
marcomuzio: :O

abigailbishop: How do I fix it? Just `del station_r` and `del station_d`? Shouldn't that cleanup be handled by the function?

araproc/framework/dataset.py
abigailbishop: Yeah `station[0]` refers to station center so I believe the .GetZ() is at some non-zero depth in ice as well. And I did test it and the vertex distribution for events generated by AraSim looks as expected.

araproc/framework/dataset.py
marcomuzio: Okay, this is annoying. I'm a little annoyed it got like this though it's probably my fault. I don't see an immediate & good solution to fixing get_vpp and get_jackknife_rms. But I do think they should both work fine if you pass a TGraph? So can you just package everything up with `framework.waveform_utilities.arrays_to_tgraph`?

araproc/framework/dataset.py
marcomuzio: I think you shouldn't work with the file directly. It looks like the `SimWrapper` class has member `self.report_ptr` which is the `Report` object. We should add the `Detector` object analogously to here https://github.com/ara-software/AraProc/blob/831d13b56e1211c86d84fbfc4fa88b3367c1d26e/araproc/framework/dataset.py#L632 but using the `self.sim_settings_tree` since that's the `AraTree`.

araproc/framework/dataset.py
```diff
@@ -73,6 +74,142 @@ def get_filters(station_id, analysis_config):
 
     return cw_filters
 
+def get_best_antenna(file):
```
marcomuzio, line 77: So this will be a member function of `SimWrapper`

araproc/framework/dataset.py
```diff
@@ -73,6 +74,142 @@ def get_filters(station_id, analysis_config):
...
+        # If this antenna's waveform is greater than the saved SNR, save this 
+        #   antennas string index, antenna index, and SNR as the best
+        if SNR > best_SNR: 
+            best_SNR = SNR
+            best_ant = (s, a)
+
+    # Return the triggering antenna with the greatest SNR
+    return best_ant
+
+def get_AraSim_xyz_position(source, target, icemodel):
```
marcomuzio, line 168: Also this should be a member function of `SimWrapper`

araproc/framework/dataset.py
abigailbishop: I set these to match the `Earth_to_Flat_same_angle` definition in AraSim. Do you still want me to change the variable names? I also suggest that target is the station and source is the neutrino in the comments because I was also turned around by the naming. 

https://github.com/ara-software/AraSim/blob/2758c07dc56fb1a9a784e460306868f0f940b499/RaySolver.cc#L39

araproc/framework/dataset.py
abigailbishop: Sure, done

araproc/framework/dataset.py
abigailbishop: Done

araproc/framework/dataset.py
abigailbishop: Done

araproc/framework/dataset.py
marcomuzio: I think it's better to be clear here, since probably many people reading this won't read that code? Especially since you don't call that function.

abigailbishop: Done, lmk what you think

araproc/framework/dataset.py
```diff
@@ -905,17 +775,144 @@ def get_sim_information(self,
...
+        x_origin = r_from_pole_origin * np.cos(lon_origin)
+        y_origin = r_from_pole_origin * np.sin(lon_origin)
+
+        # Get depth of the position
+        position_depth = self.icemodel_ptr.Surface( position.Lon(), position.Lat()) - position.R()
+        ang_diff = position.Angle(origin)
+        depth_diff = position.R() - origin.R()*np.cos(ang_diff)
+        z_origin = -position_depth - depth_diff
+
+        return x_origin-x_position, y_origin-y_position, z_origin
```
marcomuzio, line 914: lol sorry I think I'm confused about this again. Shouldn't it be `x_position-x_origin`? `origin` is the origin of the coordinate system we are transforming to isn't it?

araproc/framework/dataset.py
marcomuzio: Looks good!

araproc/framework/dataset.py
abigailbishop: You're totally right to be confused, it is was wrong but I fixed it now

marcomuzio: Thanks!

marcomuzio (approved):
All good on my end after final cross checks!

PR #127, opened 2025-05-12
Add functionality for max CW ID frequency

marcomuzio:
Adds a `max_cw_id_freq` parameter which controls the maximum frequency up to which CW ID is used to activate filters. CW filters from `min_cw_id_freq` to `max_cw_id_freq` are only activated if they cover a frequency IDed in the CW ID file. Filters outside this band are always turned on.

### DIFF
--- a/araproc/analysis/cw_filter.py
+++ b/araproc/analysis/cw_filter.py
@@ -44,7 +44,7 @@ def apply_filters_one_channel(cw_filters, active_cw_filters, waveform_in):
     
     return latest_waveform
 
-def apply_filters(cw_filters, waveform_bundle, cw_ids = None, min_cw_id_freq = 0.0):
+def apply_filters(cw_filters, waveform_bundle, cw_ids = None, min_cw_id_freq = 0.0, max_cw_id_freq = 1.0):
 
     """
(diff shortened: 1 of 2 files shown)


araproc/analysis/cw_filter.py
```diff
@@ -64,6 +64,10 @@ def apply_filters(cw_filters, waveform_bundle, cw_ids = None, min_cw_id_freq = 0
         A tuple of numpy arrays containing cw id info.
         If None, all set CW filters allowed to filter. Otherwise, only those 
         covering frequencies in the cw id info will be allowed to filter.
+    min_cw_id_freq : float
```
clark2668, line 67: Can we name this variable better please? Maybe "always_on_min_cw_id_freq? Ditto for the max below? I know it's in the docs, but I think it's otherwise confusing.

clark2668 (changes requested):
Added a request for clarity. Otherwise gtg. But please apply my naming request for all instances (I only made the request once).

clark2668:
Thank you! Good to merge.

marcomuzio:
Thanks!

PR #129, opened 2025-05-15
Changes in A4 filters

pgiri4:
Changes to adjust the new CW filtering techniques


marcomuzio (approved):
If these have all been tested on a few runs and things are looking okay, then I think it's good to go.

pgiri4:
Thank you!

PR #130, opened 2025-05-15
A3 filters

Machtay:
This PR makes some small adjustments to the A3 filters. There were some large ranges of identified CW frequencies (ex: 150-210) that I did not add a wide filter for because they did not produce skymaps and spectra that look like CW. Those events are also temporally clustered, so a ST cut should remove them.

### DIFF
--- a/araproc/framework/config_files/analysis_configs.yaml
+++ b/araproc/framework/config_files/analysis_configs.yaml
@@ -194,8 +194,8 @@ station3:
     filters:
       filt1 : { min_freq : 0.000, max_freq : 0.120, min_power_ratio : 0.10 }
       filt2 : { min_freq : 0.120, max_freq : 0.128, min_power_ratio : 0.02 }
-      filt3 : { min_freq : 0.128, max_freq : 0.145, min_power_ratio : 0.10 }
-      filt4 : { min_freq : 0.145, max_freq : 0.155, min_power_ratio : 0.02 }
+      filt3 : { min_freq : 0.128, max_freq : 0.140, min_power_ratio : 0.10 }
+      filt4 : { min_freq : 0.140, max_freq : 0.155, min_power_ratio : 0.02 }
       filt5 : { min_freq : 0.155, max_freq : 0.245, min_power_ratio : 0.10 } 
       filt6 : { min_freq : 0.245, max_freq : 0.255, min_power_ratio : 0.02 }
       filt7 : { min_freq : 0.255, max_freq : 0.295, min_power_ratio : 0.10 }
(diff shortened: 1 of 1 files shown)


araproc/framework/config_files/analysis_configs.yaml
```diff
@@ -337,21 +346,24 @@ station3:
...
+      filt11 : { min_freq : 0.305, max_freq : 0.350, min_power_ratio : 0.10 }
+      filt12 : { min_freq : 0.350, max_freq : 0.400, min_power_ratio : 0.10 }
+      filt13 : { min_freq : 0.400, max_freq : 0.410, min_power_ratio : 0.02 }
+      filt14 : { min_freq : 0.410, max_freq : 0.430, min_power_ratio : 0.10 }
+      filt15 : { min_freq : 0.430, max_freq : 0.470, min_power_ratio : 0.01 }
+      filt16 : { min_freq : 0.470, max_freq : 0.495, min_power_ratio : 0.10 }
+      filt17 : { min_freq : 0.495, max_freq : 0.510, min_power_ratio : 0.01 }
+      filt18 : { min_freq : 0.510, max_freq : 0.580, min_power_ratio : 0.10 }
+      filt19 : { min_freq : 0.580, max_freq : 0.620, min_power_ratio : 0.01 }
+      filt20 : { min_freq : 0.620, max_freq : 1.000, min_power_ratio : 0.10 }
```
marcomuzio, line 366: Did you spot check running L2 that this many filters is still tractable? Or does it really slow things down?

marcomuzio (commented):
One small comment, then probably good to go

araproc/framework/config_files/analysis_configs.yaml
Machtay: Ah I didn't. I'm not sure the last time I specifically ran L2 so I don't have a sense of what to compare to, but maybe I can try on a sample of events where it looks like we have CW (or at least, where we have a lot of frequencies IDed as CW)?

marcomuzio: That will probably be the worst case scenario. Maybe compare it to another station that has fewer filters?

marcomuzio: Or even another A3 config?

Machtay: Just tested it on another config in A3 with Paramita and it looks like the same speed.

marcomuzio: What's that speed per event roughly?

Machtay: Some were 0.75 some were about 1.0

drparamita: agree. We just checked with 2 configs and the speed per event is not significantly different.

marcomuzio: Great!

marcomuzio (approved):

PR #131, opened 2025-05-15
A1 Config 1 and 2 updates

pgiri4:
I might need a few more hours to make changes to config 3, but I will try to do it as soon as possible.


marcomuzio (approved):
As long as this was checked on some test runs, I think it looks good!

PR #132, opened 2025-05-15
A5 C1 Filters

Machtay:
Added in some filters for C1. These largely look the same as the ones for C2.

### DIFF
--- a/araproc/framework/config_files/analysis_configs.yaml
+++ b/araproc/framework/config_files/analysis_configs.yaml
@@ -468,16 +468,18 @@ station5:
     excluded_channels : []
     filters:
       filt1 : { min_freq : 0.000, max_freq : 0.045, min_power_ratio : 0.10 }
-      filt2 : { min_freq : 0.045, max_freq : 0.065, min_power_ratio : 0.01 }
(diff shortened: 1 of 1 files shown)


araproc/framework/config_files/analysis_configs.yaml
drparamita, line 1: filt12 : { min_freq : 0.750, max_freq : 0.756, min_power_ratio : 0.0005 } band is narrow. 
Are you sure about the band is so narrow? How many frequencies you see in this range? This filter is very strong.

drparamita: I am sure you cross-checked the power threshold. I wanted to be sure about the bandwidth here. Thanks

marcomuzio (approved):
Looks good to me.

araproc/framework/config_files/analysis_configs.yaml
Machtay: Ah yeah, this is left over from before. I think I only saw peaks very narrowly in this band consistently.

drparamita: ok, thanks. 
Looks good to me. You can go ahead and merge, I think.

PR #135, opened 2025-05-22
update the dedisperser with padding

clark2668:
This updates the dedisperser module. The idea here is that we want to pad the waveform with some buffer (in this case, 40 ns) before the dedispersion. This is necessary because the group delay of the system pushes the signal forward, so when it is removed, the signal shifts backwards in time. If we don't have any pre-padding, then the signal wraps around to the end of the trace, which is explicitly incorrect. We don't want to keep all of the pre-pad tho. So in this case, we pre-pad by 40 ns, and then keep 10 ns. The 10 ns is chosen by studying the largest group delay we see in the system response where the bandpass is still substantial. See the following plot.

channel_0_gaussian_pulse

### DIFF
--- a/araproc/analysis/dedisperse.py
+++ b/araproc/analysis/dedisperse.py
@@ -54,7 +54,7 @@ def dedisperse_wave(
         volts, # in volts,
         phase_spline, # the phase spline
         pre_pad_ns=40.0,  # amount of zero prepadding before dedispersion (ns)
-        final_crop_ns=10.0  # amount of padding to keep after dedispersion (ns)
+        final_crop_ns=10.0  # amount of prepadding to keep after dedispersion (ns)
         ):
     
     """
--- a/araproc/analysis/dedisperse.py
+++ b/araproc/analysis/dedisperse.py
@@ -53,7 +53,7 @@ def dedisperse_wave(
         times, # in nanoseconds,
         volts, # in volts,
         phase_spline, # the phase spline
-        pre_pad_ns=40.0,  # amount of zero padding before dedispersion (ns)
+        pre_pad_ns=40.0,  # amount of zero prepadding before dedispersion (ns)
         final_crop_ns=10.0  # amount of padding to keep after dedispersion (ns)
         ):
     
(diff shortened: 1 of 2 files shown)


araproc/analysis/dedisperse.py
```diff
@@ -82,20 +93,44 @@ def dedisperse_wave(
...
 
-    # first thing to do is get the frequency domain representation of the trace
-
-
-    freqs, spectrum = wu.time2freq(times, volts)
+    # Calculate time step from the input times array
+    dt = times[1] - times[0]  # assuming uniform sampling
+    
+    # Calculate number of padding samples needed
+    n_pad_samples = int(pre_pad_ns / dt)
```
marcomuzio, line 100: Does this need to be a round?

araproc/analysis/dedisperse.py
```diff
@@ -82,20 +93,44 @@ def dedisperse_wave(
...
+    
+    # Create padded time array: from (t_i - pre_pad_ns) to t_f
+    t_start_padded = times[0] - pre_pad_ns
+    padded_times = np.arange(t_start_padded, times[-1] + dt/2, dt)
+    
+    # Create padded voltage array with zeros at the beginning
+    padded_volts = np.zeros(len(padded_times))
+    padded_volts[n_pad_samples:n_pad_samples+len(volts)] = volts
```
marcomuzio, line 108: why do we need this range? shouldn't it be the right length to just do `padded_volts[n_pad_samples:]`?

araproc/analysis/dedisperse.py
```diff
@@ -82,20 +93,44 @@ def dedisperse_wave(
...
     # do complex division to do the dedispersion
     spectrum /= phased_rewrapped
 
     # back to the time domain
-    times, volts = wu.freq2time(times, spectrum)
-    return times, volts
+    dedispersed_times, dedispersed_volts = wu.freq2time(padded_times, spectrum)
+    
+    # Crop the result: keep from (t_i - final_crop_ns) to t_f
+    t_start_final = times[0] - final_crop_ns
```
marcomuzio, line 126: this is a confusing name

marcomuzio (changes requested):

clark2668:
Take another look now?

PR #137, opened 2025-05-23
swap in the A1 when grepping around for sensor hk

clark2668:
This sensor hk file utility finder needs to search for A1, not A100.

### DIFF
--- a/araproc/framework/file_utilities.py
+++ b/araproc/framework/file_utilities.py
@@ -45,14 +45,17 @@ def guess_location_of_hk_files(station_id, run_number):
     You need to be at the WIPAC datawarehosue for this to work.
     """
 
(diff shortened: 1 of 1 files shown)


araproc/framework/file_utilities.py
```diff
@@ -45,14 +45,17 @@ def guess_location_of_hk_files(station_id, run_number):
     You need to be at the WIPAC datawarehosue for this to work.
     """
 
-    inpath = f"/data/exp/ARA/*/L1/100pct/ARA0{station_id}/*/run{run_number:06d}/"
+    if station_id==100:
+        local_station_id=1
```
marcomuzio, line 49: set for other stations

marcomuzio (changes requested):

araproc/framework/file_utilities.py
clark2668: Done. Look good now?

PR #138, opened 2025-05-23
Typo in readout limits made it not work.

Machtay:
I guess I should let you read this one before badgering you to let me merge, Marco


marcomuzio (approved):
fine.

PR #145, opened 2025-05-24
Cw redo a5

Machtay:
This PR consolidates filters on A5. The goal is to have the less harsh filters be broader and make the harshest filters narrow (and possibly subsumed by the broad filters).


marcomuzio (approved):
Looks good, nice job!

PR #154, opened 2025-06-02
Add new A2 config

abigailbishop:
The new config will appear to split config 7 into two so it has the same channels masked.


marcomuzio (approved):
looks good to me

PR #164, opened 2025-09-10
Updated CW removal for remaining high corr -- low snr events.

Machtay:
This contains the updated CW filters I implemented to remove the last few CW events in the low snr -- high correlation tail.


marcomuzio (changes requested):
Thanks @Machtay! I think this just needs the comments to be cleaned up. And what about config2? Does that need an update?

PR #166, opened 2025-10-01
Grey out masked channels

abigailbishop:
Add an argument to our `framework.data_visualization.plot_waveform_bundle()` function that lists a series of antennas to plot in a slightly grayed out color with the annotation "Masked to Analysis". For example if someone were to pass `excluded_channels=framework.dataset.AnalysisDataset.excluded_channels`, all channels masked to analysis would be plotted in greyed colors to help communicate to analyzers that the antennas did not contribute to analysis variable generation. Examples shown below: 

[station_2_run_18880_eventNo_47220_spectrum.pdf](https://github.com/user-attachments/files/22648500/station_2_run_18880_eventNo_47220_spectrum.pdf)
[station_2_run_18880_eventNo_47220_trace.pdf](https://github.com/user-attachments/files/22648505/station_2_run_18880_eventNo_47220_trace.pdf)

### DIFF
--- a/araproc/framework/data_visualization.py
+++ b/araproc/framework/data_visualization.py
@@ -1,5 +1,6 @@
 import numpy as np
 import matplotlib.pyplot as plt
+from araproc.framework import constants as const
 from araproc.framework import waveform_utilities as wu
 from araproc.framework import map_utilities as mu
 import ROOT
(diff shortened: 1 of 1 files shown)


araproc/framework/data_visualization.py
```diff
@@ -17,10 +17,29 @@
...
 matplotlib.use("pdf")
 
+def get_wf_color(channel_number, excluded_channels): 
+    """
+    Given a channel number and a list of excluded channels, return a purple 
+      color for all VPols (channel_number<8) in purple, a green color for
+      HPols (channel_number > 8), and a slightly greyer purple or green if the 
+      channel is masked.
+    """
+    if channel_number < 8: # Vpols
```
marcomuzio, line 27: Might be better to use the constant containing the vpol channels and check if the number is in there.

abigailbishop: Should be good to go now!

marcomuzio (approved):

PR #167, opened 2025-10-03
Add is_noise flag to sim_info

marcomuzio:


abigailbishop:
If you tested it and it worked correctly for signal and noise sims then I'm happy with this.

marcomuzio:
Yes, it seemed to work in tests! So will merge. Thanks!

PR #197, opened 2026-06-05
Changed ray trace table dir and included systematics in standard reco.

Machtay:
Sets a new default directory for the ray trace tables. This also adds an argument for the standard_reco class for the systematics, so in fivestation we'll need to specify 41 (up) or 42 (down) if we want to run with the ice model systematics on.

### DIFF
--- a/araproc/analysis/standard_reco.py
+++ b/araproc/analysis/standard_reco.py
@@ -40,7 +40,7 @@ class StandardReco:
         The list of channels you want *exclued* from the interferometry
     station_id : int
         The id of the station you want to reco (only station 1-5 supported)
-    systematics : int
+    icemodel : int
         Which ray trace tables to be used. 40 is nominal, 41 is up, 42 is down.
     requested_maps : set of strings
         The keys for the reco maps to be generated.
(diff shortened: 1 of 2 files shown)


araproc/analysis/standard_reco.py
```diff
@@ -40,6 +40,8 @@ class StandardReco:
         The list of channels you want *exclued* from the interferometry
     station_id : int
         The id of the station you want to reco (only station 1-5 supported)
+    systematics : int
```
clark2668: I'd prefer this wasn't called "systematics" and instead just called it "icemodel". It seems nitpicky, but I think it will help with clarity.

clark2668 (changes requested):
Small request

araproc/analysis/standard_reco.py
Machtay: done

marcomuzio (approved):
Lgtm

araproc/analysis/standard_reco.py
clark2668: Awesome

clark2668 (approved):

PR #199, opened 2026-07-06
make function for best triggering antenna

Machtay:
This separates out the meat of get_best_antenna so we can use the logic in other instances. Motivated for this PR: https://github.com/ara-software/FiveStation/pull/608

### DIFF
--- a/araproc/framework/dataset.py
+++ b/araproc/framework/dataset.py
@@ -1036,11 +1036,6 @@ class SimWrapper:
             objects, respectively
         """
 
-        # Delegate to the module-level implementation so that other callers
-        # (e.g. fivestation's veff_calculator, which needs to do this same
-        # lookup directly from bare ROOT.TTreeReader objects rather than a
-        # SimWrapper instance) can share the exact same logic instead of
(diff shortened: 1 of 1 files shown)


araproc/framework/dataset.py
```diff
@@ -948,77 +1036,12 @@ def get_best_antenna(self):
...
-  
-            # If this antenna's waveform is greater than the saved SNR, save this 
-            #   antennas string index, antenna index, and SNR as the best
-            if SNR > best_SNR: 
-                best_SNR = SNR
-                best_ant = (s, a)
-
-        # Return the triggering antenna with the greatest SNR
-        return best_ant
+        # Delegate to the module-level implementation so that other callers
```
marcomuzio: We can probably drop this explanation here

marcomuzio (approved):
Thanks for doing this refactor Alex! Just one small comment but I think this is good to merge.

araproc/framework/dataset.py
Machtay: Sweet, changes

PR #200, opened 2026-07-13
Add trig receipt

Machtay:
This adds a helper function for getting the average receipt angle of an event for the antennas that triggered.

### DIFF
--- a/araproc/framework/dataset.py
+++ b/araproc/framework/dataset.py
@@ -164,14 +164,13 @@ def find_best_triggering_antenna(report_station, detector_station):
     # return the triggering antenna with the greatest SNR
     return best_ant
 
-def find_avg_receipt_ang(report_station, detector_station):
+def find_avg_receipt_ang(report_station):
     """
     Calculate the average receiving angle of the signal by the antennas. 
 
     Parameters
     ----------
     report_station : ROOT AraSim Report::StationReport object (i.e. report.stations[0])
-    detector_station : ROOT AraSim Detector::ARA_station object (i.e. detector.stations[0])
 
     Returns
     -------
@@ -179,20 +178,8 @@ def find_avg_receipt_ang(report_station, detector_station):
         Average zenith angle (measured from nadir = 0) of signal seen by triggered antennas.
     """
 
-    # find all antennas that participated in the trigger
-    trig_ants = [
-        (s, a)
-        for s in range(len(report_station.strings))
-        for a in range(len(report_station.strings[s].antennas))
-        if report_station.strings[s].antennas[a].Trig_Pass
-    ]
-
-    # if the event doesn't trigger, put a sentinel to mark it
-    # a very negative value will also fail any rec ang cut so it isn't mistakenly counted
-    if len(trig_ants) == 0:
-        return -1000 
-
-    avg_rec_ang = np.mean(report_station.strings[s].antennas[a].theta_rec for s, a in trig_ants)
+    rec_angs = [report_station.strings[s].antennas[a].theta_rec for s in report_station.strings for a in report_station.strings[s].antennas[a]]
+    avg_rec_ang = np.mean(rec_angs)
 
     return avg_rec_ang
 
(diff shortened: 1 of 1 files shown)


araproc/framework/dataset.py
```diff
@@ -164,6 +164,37 @@ def find_best_triggering_antenna(report_station, detector_station):
...
+    report_station : ROOT AraSim Report::StationReport object (i.e. report.stations[0])
+    detector_station : ROOT AraSim Detector::ARA_station object (i.e. detector.stations[0])
+
+    Returns
+    -------
+    avg_rec_ang : float
+        Average zenith angle (measured from nadir = 0) of signal seen by triggered antennas.
+    """
+
+    # find all antennas that participated in the trigger
```
marcomuzio: I see the logic here, but I wonder if this will cause a slight mismatch with the definition @Mohammad-Neutrino used? I think he just took a simple average over all channels?

araproc/framework/dataset.py
```diff
@@ -164,6 +164,37 @@ def find_best_triggering_antenna(report_station, detector_station):
...
+    trig_ants = [
+        (s, a)
+        for s in range(len(report_station.strings))
+        for a in range(len(report_station.strings[s].antennas))
+        if report_station.strings[s].antennas[a].Trig_Pass
+    ]
+
+    # if the event doesn't trigger, put a sentinel to mark it
```
marcomuzio: Can this ever happen? Generally, the average receipt angle seems well-defined regardless of the trigger outcome on the channel? Maybe the only exception is if there's no ray-trace solution to the channel, what happens then?

araproc/framework/dataset.py
```diff
@@ -164,6 +164,37 @@ def find_best_triggering_antenna(report_station, detector_station):
...
+    # a very negative value will also fail any rec ang cut so it isn't mistakenly counted
+    if len(trig_ants) == 0:
+        return -1000 
+
+    avg_rec_ang = np.mean(report_station.strings[s].antennas[a].theta_rec for s, a in trig_ants)
```
marcomuzio: maybe it makes more sense to store the receipt angle directly rather than the `(s,a)` tuple?

araproc/framework/dataset.py
```diff
@@ -164,6 +164,37 @@ def find_best_triggering_antenna(report_station, detector_station):
...
+
+    Parameters
+    ----------
+    report_station : ROOT AraSim Report::StationReport object (i.e. report.stations[0])
+    detector_station : ROOT AraSim Detector::ARA_station object (i.e. detector.stations[0])
+
+    Returns
+    -------
+    avg_rec_ang : float
+        Average zenith angle (measured from nadir = 0) of signal seen by triggered antennas.
```
marcomuzio: This angle is stored in degrees in AraSim, right? Just to be sure

marcomuzio (changes requested):
Just a few small things

araproc/framework/dataset.py
Machtay: I guess I figured we wouldn't want to muddy the reconstruction with those antennas, but yeah if that's what Mohammad did then I can use them all

araproc/framework/dataset.py
Machtay: Yeah I wasn't sure so I put this in just in case. I agree in data I assume not, but I thought maybe in sims it could be recorded and we wouldn't want to count it.

araproc/framework/dataset.py
Machtay: Actually I don't think so. The comment here in AraSim doesn't say, but rec_ang nearby is in radians: https://github.com/ara-software/AraSim/blob/4fcf1f373a021677e25b5b5f8e7ae9ae56f409e8/Report.h#L63

Also here it looks like it's converted to degrees when we check the gain, so I think it's radians: https://github.com/ara-software/AraSim/blob/4fcf1f373a021677e25b5b5f8e7ae9ae56f409e8/Report.cc#L1007

araproc/framework/dataset.py
Machtay: Cleaned this up

araproc/framework/dataset.py
Machtay: Ok, Mohammad confirmed it was all antennas, so I simplified this

araproc/framework/dataset.py
marcomuzio: Okay, we should make clear the units that this function returns in. Especially because the units this returns and the units of the thresholds need to agree.

marcomuzio (changes requested):
Just one small thing, then it's good to go.

marcomuzio (approved):

PR #201, opened 2026-07-22
fix bug in getting avg rec ang

Machtay:
The previous version of this function was actually only checking across all interactions. This should now pick out the particular one, so we get an average for the particular event we are checking.

### DIFF
--- a/araproc/framework/dataset.py
+++ b/araproc/framework/dataset.py
@@ -5,6 +5,7 @@ import os
 import numpy as np
 import ROOT
 import yaml
+import warnings
 from scipy import stats
 
 
--- a/araproc/framework/dataset.py
+++ b/araproc/framework/dataset.py
@@ -1049,7 +1049,7 @@ class SimWrapper:
         sim_info["is_noise"] = (int(self.settings_ptr.TRIG_ANALYSIS_MODE) == 2) # modes 0 & 1 are for signal, 2 is pure noise
 
         # get the receipt angle for the triggered antennas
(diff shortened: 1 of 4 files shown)


araproc/framework/dataset.py
```diff
@@ -164,22 +164,33 @@ def find_best_triggering_antenna(report_station, detector_station):
...
     """
 
-    rec_angs = [report_station.strings[s].antennas[a].theta_rec for s in report_station.strings for a in report_station.strings[s].antennas[a]]
-    avg_rec_ang = np.mean(rec_angs)
+    # Get the received angles for that interaction
+    rec_angs = [
+        float(rec_ang)
+        for string in report_station.strings
+        for antenna in string.antennas
+        if interaction_idx < len(antenna.theta_rec)
```
marcomuzio: I feel like if this is violated we should throw an error rather than return nan. Could be here or below.

marcomuzio (changes requested):
One small thing. Thanks for fixing this

araproc/framework/dataset.py
Machtay: I can do that, but then don't we want it to return something in case that happens so it doesn't totally disrupt either the processing or the veff calculation?

marcomuzio: Does this ever happen for the Veff calculation? If it does I feel like something must be very wrong and we want to flag it

clark2668:
Okay, I sort of hate that this requires calling so much C++, but so be it. I think this can be merged.

marcomuzio:
Ya, I'm not sure how to avoid it without reimplementing the ray tracer unfortunately.

PR #205, opened 2026-07-31
Pretty waveforms

clark2668:
Pretty waveforms for paper

### DIFF
--- a/araproc/framework/data_visualization.py
+++ b/araproc/framework/data_visualization.py
@@ -346,18 +346,11 @@ def plot_waveform_bundle_pretty(
     mpl.rcParams.update({'font.size':28})
     mpl.rcParams['font.family'] = 'STIXGeneral'
     mpl.rcParams['mathtext.fontset'] = 'stix'
-    # plt.rcParams.update({
-    #     "axes.titlesize": 28,
-    #     "axes.labelsize": 28,
-    #     "xtick.labelsize": 28,
-    #     "ytick.labelsize": 28,
-    #     "legend.fontsize": 28
-    # })
 
     fig, axes = plt.subplots(
-        4, 4, figsize=(12, 10), sharex=True, sharey=True, # layout="constrained",
+        4, 4, figsize=(20, 10), sharex=True, sharey=True, # layout="constrained",
     )
-    fig.subplots_adjust(wspace=0, hspace=0, left=0.09, right=0.98, top=0.98, bottom=0.08)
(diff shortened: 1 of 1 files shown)


araproc/framework/data_visualization.py
```diff
@@ -275,6 +275,121 @@ def plot_waveform_bundle(
...
+    for row, row_channels in enumerate(_PRETTY_CHANNEL_LAYOUT):
+        for col, ch in enumerate(row_channels):
+            ax = axes[row, col]
+            if ch not in waveform_dict:
+                ax.axis("off")
+                continue
+
+            times, volts = wu.tgraph_to_arrays(waveform_dict[ch])  # times in ns, volts in mV
+            ax.axhline(0, color="lightgray", linestyle="--", linewidth=1, zorder=0)
+            ax.plot(times, volts/1000., color=get_wf_color(ch, excluded_channels), linewidth=1.5) # millivolts
```
Mohammad-Neutrino, line 364: Can we keep default volts in mV?

araproc/framework/data_visualization.py
```diff
@@ -275,6 +275,121 @@ def plot_waveform_bundle(
...
+                0.97, 0.95, f"Channel {ch}",
+                transform=ax.transAxes, ha="right", va="top", fontsize=26,
+            )
+            if ch in excluded_channels:
+                ax.annotate(
+                    "Masked to Analysis", (0.03, 0.87),
+                    xycoords="axes fraction", color="grey", fontsize=9,
+                )
+
+            if xlim is not None:
```
Mohammad-Neutrino, line 377: Also, the buffer before and after the waveform makes them look squeezed. Maybe reduce that or keep min(16 traces) to max(16 traces).

Mohammad-Neutrino (commented):
This looks real pretty! Just a couple of minor comments. Thanks!

marcomuzio (approved):

araproc/framework/data_visualization.py
clark2668: So, the challenge with this trace is that its so blown out. So it goes up to like +-2.5V. So if I leave it in mV, then I have to waste lots of space on the left side with numbers like 1000. So I think for this specific plot let's go with volts?

araproc/framework/data_visualization.py
clark2668: So, I switched to a more horizontal layout in the final version of this that I think makes them look significantly less squeezed!

